PR #1, opened 2023-01-13
Add project for review 1.0

kimfom01:


kimfom01.Console.MathGame/MathGame/Controllers/GameController.cs
```diff
@@ -0,0 +1,134 @@
+﻿using MathGame.UserInput;
+
+namespace MathGame.Services;
+
+public class GameController
+{
+    private Input _input = new Input();
+    private Dictionary<string, string> _additionQuestions = new()
+    {
+        { "5 + 7", "12" },
```
TheCSharpAcademy, line 10: Please add random numbers from 1 to 10 so each game is different. 

I'll tell you in advance that the logic for division needs to have random number 1 to 100 and the result needs to be a full number. (I'll add that to the requirements).

TheCSharpAcademy (changes requested):

kimfom01.Console.MathGame/MathGame/Controllers/GameController.cs
kimfom01: yeah I think this is better than hard coding the questions. I will work on that. I think this will require me to have a calculator class to perform the calculations and validate the answers. not gonna lie I started out this way but scrapped it because I thought hard coding the questions was an easier approach 😂😂

kimfom01: Hmmm I think I need to think this one through 🤣

TheCSharpAcademy: @kimfom01 you've got this! 😊

TheCSharpAcademy: @kimfom01 Remember I made a video on this if you need help. But you should be able to do it on your own 😁https://www.youtube.com/watch?v=YUc5NZZVtjA&t=16s

kimfom01: I'll finish it after my TOEFL exams 😅

kimfom01.Console.MathGame/MathGame/Controllers/GameController.cs
```diff
@@ -0,0 +1,137 @@
...
+    {
+        int count = 0, score = 0;
+
+        for (int i = 0; i < 10; i++)
+        {
+            var numA = Random.Shared.Next(1, 10);
+            var numB = Random.Shared.Next(1, 10);
+            if (operation == Operation.Division)
+            {
+                numA = Random.Shared.Next(10, 100);
```
TheCSharpAcademy, line 88: hey! we're almost there! sorry if the requirements were confusing, but you should only allow operations where the division results in a full number.  You need some sort of loop that will keep running if the equation is 27/6 for example. 
image

TheCSharpAcademy (changes requested):

kimfom01.Console.MathGame/MathGame/Controllers/GameController.cs
kimfom01: so sort of like the game shouldn't present questions that could lead to answers with decimal point right?

PR #25, opened 2023-03-09
added my project

lucasnates:


lucasnates.MathGame/ConsoleMathGame/GameEngine.cs
```diff
@@ -0,0 +1,275 @@
+﻿using ConsoleMathGame.Models;
+
+namespace ConsoleMathGame
```
kimfom01, line 3: Just like you did in your Game.cs model class, in your next projects you should use file scoped namespaces to reduce number of indentation 😉

lucasnates.MathGame/ConsoleMathGame/Helpers.cs
kimfom01, line 1: "Separation of Concerns" is a programming principle that encourages you to organize your code into classes that handle similar operations, rather than creating a "Helper" class with unrelated methods. This can make your code more organized, easier to read, and simpler to maintain. I'd like you to do a little research and learn about this 😊

lucasnates.MathGame/ConsoleMathGame/Menu.cs
```diff
@@ -0,0 +1,70 @@
+﻿namespace ConsoleMathGame
+{
+    internal class Menu
+    {
+        GameEngine gameEngine = new();
+        internal static bool isDivisionOn;
```
kimfom01, line 6: It's a good practice to minimize the amount of shared state in your program and use encapsulation to keep data private and controlled. For this project it might not be an issue to have static fields but as your project size increases, it creates room for bugs that are hard to determine what went wrong 😊

kimfom01 (approved):

lucasnates:
Thank you @kimfom01 !!! I will learn from this 😄

PR #26, opened 2023-03-11
finished

jgone99:


TheCSharpAcademy:
@jgone99 Hi there! Well done on finishing the project! Could you please submit the link to this PR via the dashboard? This way I can automatically tick it off once it's been reviewed. 🤓You can just click on 'Submit Project' in the dashboard as per screenshot: 
222241698-3503a653-f383-410b-88ef-20ee0ba0ac9a

TheCSharpAcademy (changes requested):
Hi there! Again well done, thanks for sending your code for review!

It looks like you didn't submit your .csproj file (and .sln if you had one). I was expecting to see a folder structure similar to this: 

image

TheCSharpAcademy:
@jgone99 hey there! Just checking if you need any help?

TheCSharpAcademy:
@jgone99 Hi there! I'm finding the idea very original, but I can't understand what happens after the first equation. Could you explain what happens based on this example? 

image

jgone99:
The (A) operand always takes the result of the previous operation. That is why the (RESULT) and (A) are always the same. Operand (B) is always a new, randomly generated integer. Each round, the user must choose which of the 4 arithmetic operations to apply to the two operands with the goal of ultimately getting the (RESULT) within a certain error margin of the (TARGET). In this particular example, (A = 4) and (B = 18). Suppose the user chooses (+), then (RESULT = 4 + 18 = 22). Again, the next (A) will be the same as the new (RESULT) so (A = 22) for the next round. It's not the most exciting game since the outcome is mostly governed by chance but as far as I could tell it satisfied the requirements.

TheCSharpAcademy:
> The (A) operand always takes the result of the previous operation. That is why the (RESULT) and (A) are always the same. Operand (B) is always a new, randomly generated integer. Each round, the user must choose which of the 4 arithmetic operations to apply to the two operands with the goal of ultimately getting the (RESULT) within a certain error margin of the (TARGET). In this particular example, (A = 4) and (B = 18). Suppose the user chooses (+), then (RESULT = 4 + 18 = 22). Again, the next (A) will be the same as the new (RESULT) so (A = 22) for the next round. It's not the most exciting game since the outcome is mostly governed by chance but as far as I could tell it satisfied the requirements.

I didn't exactly say what the game was supposed to entail but I was expecting something more simple, such as simply verifying the if the users inputs the correct result for the operation 😁. However I'll approve it anyways, given the originality. 😁 Looking forward to see your work in the next projects!

TheCSharpAcademy:
@jgone99 Also, not sure you care about this, but could you please mark the intro section's articles as read so I can upgrade your belt to green? 😁

PR #28, opened 2023-03-14
MathGame

JsPeanut:


MathGame.JsPeanut/MathGame.JsPeanut/Program.cs
```diff
@@ -0,0 +1,320 @@
+﻿using System;
```
edvaudin, line 1: Removing unnecessary using statement.

MathGame.JsPeanut/MathGame.JsPeanut/Program.cs
```diff
@@ -0,0 +1,320 @@
...
+
+class Program
+{
+
+    public static void Main(string[] args)
+    {
+        List<Operation> previousGames = new List<Operation>();
```
edvaudin, line 8: We can shorten this to the following:
```suggestion
        List<Operation> previousGames = new();
```

MathGame.JsPeanut/MathGame.JsPeanut/Program.cs
```diff
@@ -0,0 +1,320 @@
...
+
+        bool doUserStillWantToPlay = true;
+
+        Console.WriteLine("Hello there! Welcome to this math game.");
+        Console.WriteLine("Available commands: -Score -Menu");
+        Console.WriteLine("Do you want to play?");
+        Console.WriteLine("Type 'yes' if so. If not, type 'exit'.");
```
edvaudin, line 15: See if you can reduce this down to one Console.WriteLine() call using string literals.

MathGame.JsPeanut/MathGame.JsPeanut/Program.cs
```diff
@@ -0,0 +1,320 @@
...
+        {
+            switch (Console.ReadLine())
+            {
+                case "yes":
+                    StartGame();
+                    break;
+                case "exit":
+                    doUserStillWantToPlay = false;
+                    break;
+            }
```
edvaudin, line 27: Currently, if the user types anything other than yes or exit, nothing happens. It might be worth adding a default case so that you can remind the user that they need to either enter yes or exit.

MathGame.JsPeanut/MathGame.JsPeanut/Program.cs
```diff
@@ -0,0 +1,320 @@
...
+        }
+
+        void StartGame()
+        {
+            Random rnd = new Random();
```
edvaudin, line 32: We can shorten this to the following:
```suggestion
            Random rnd = new();
```

MathGame.JsPeanut/MathGame.JsPeanut/Program.cs
```diff
@@ -0,0 +1,320 @@
...
+
+            int Multiply(int x, int y)
+            {
+                return x * y;
+            }
+
+            int Divide(int x, int y)
+            {
+                return x / y;
+            }
```
edvaudin, line 54: I am not sure it is strictly necessary to create methods for these. We know that to do `x + y` we are adding for instance. So this extraction does not add value or any further clarity than needed.

MathGame.JsPeanut/MathGame.JsPeanut/Program.cs
```diff
@@ -0,0 +1,320 @@
...
+                Console.WriteLine();
+                Console.WriteLine("Addition (Type A)");
+                Console.WriteLine("Substraction (Type S)");
+                Console.WriteLine("Multiplication (Type M)");
+                Console.WriteLine("Division (Type D)");
+            }
+
+            void AdditionGame()
+            {
+                Operation sum = new Operation();
```
edvaudin, line 70: We can shorten this to the following: 

```suggestion
                Operation sum = new();
```

MathGame.JsPeanut/MathGame.JsPeanut/Program.cs
```diff
@@ -0,0 +1,320 @@
...
+                if (Console.ReadLine() == $"{sum.Result}")
+                {
+                    Console.WriteLine("Nice.");
+                    sum.Score++;
+                    UpdateStats(sum);
+                    do
+                    {
+                        a = rnd.Next(1, 100);
+                        b = rnd.Next(1, 100);
+                    } while (a + b == sum.Calc[0]);
```
edvaudin, line 84: Since `Calc` is a string value, I think this is only looking at the first character and trying to compare it to an integer?

I know what it is trying to do, but can you think of a way to make it clearer what this do while loop is doing?

MathGame.JsPeanut/MathGame.JsPeanut/Program.cs
```diff
@@ -0,0 +1,320 @@
...
+                    AdditionGame();
+                }
+                else
+                {
+                    Console.WriteLine("Wrong. If you typed a command, ignore this message.");
```
edvaudin, line 89: Consider having a separate check for if a user has entered a command instead of an answer so that you don't have to awkwardly respond with "Wrong" first. I also note that it takes away score regardless.

MathGame.JsPeanut/MathGame.JsPeanut/Program.cs
```diff
@@ -0,0 +1,320 @@
...
+            }
+
+            void UpdateStats(Operation operation)
+            {
+                previousGames.Add(new Operation
+                {
+                    Score = operation.Score,
+                    Result = operation.Result,
+                    Calc = operation.Calc
+                });
```
edvaudin, line 208: Why do we need to reset all the values if we can just do this instead?
```suggestion
                previousGames.Add(operation);
```

MathGame.JsPeanut/MathGame.JsPeanut/Program.cs
```diff
@@ -0,0 +1,320 @@
...
+                    }
+                }
+                else
+                {
+                    Console.WriteLine("You don't have any score.");
+                }
+
+            }
+
+            Play();
```
edvaudin, line 231: This call was a bit hard to find amongst the nested methods. Consider incorporating this into a core game loop towards the top of the class.

MathGame.JsPeanut/MathGame.JsPeanut/Program.cs
```diff
@@ -0,0 +1,320 @@
...
+            }
+
+            void PlayWithoutIntro()
+            {
+                Console.WriteLine("Choose a math operation");
+                Console.WriteLine();
+                Console.WriteLine("Addition (Type A)");
+                Console.WriteLine("Substraction (Type S)");
+                Console.WriteLine("Multiplication (Type M)");
+                Console.WriteLine("Division (Type D)");
```
edvaudin, line 270: It would be useful to include a reminder about how to access the commands here too.

MathGame.JsPeanut/MathGame.JsPeanut/Program.cs
```diff
@@ -0,0 +1,320 @@
...
+                    doUserStillWantToPlay = false;
+                    break;
+            }
+        }
+
+        void StartGame()
+        {
+            Random rnd = new Random();
+            int a = rnd.Next(1, 100);
+            int b = rnd.Next(1, 100);
```
edvaudin, line 34: I note that each game currently manages randomising the numbers itself and performs any necessary validation on it. In  that case, we possibility do not need to do it here as well.

edvaudin (changes requested):
@JsPeanut Well done on submitting your Math Game! I liked the use of the Operation class to help store related data and the program didn't throw any errors for me.

Beyond the changes I have requested in my code, consider how the program flows. There were several points when trying to use the program when I wasn't sure how to use the commands or when. Some of the instructions may need to be clearer and I think there should be some more strict validation. I liked the idea of having commands, but as I say, it was sometimes difficult to understand when I should use them.

I note that most of the logic was built into the Main() method in Program. A good habit to get into early would be to properly separate out each element of the program. One challenge I put to each of the people that submit these Math Game projects is to see how you can reduce repitition. Notice that each of your game methods do a few things in common:

- Set a calculation
- Collect user input
- Validate user input and compare with answer
- Award points if answer is correct

We could split this up into more methods so that each stage is clearly labelled. Furhtermore, it is possible that there only needs to be one game logic method, since all of the games follow the same core logic just with math operations swapped in and out.

When looking over my feedback within the code, please note that the changes/suggestions I have made in the AdditionGame() method should apply to all the others, but in any case, I hope you take the opportunity to explore combining all of the games into one method.

I hope this feedback is helpful, if you have any questions, please feel free to let me know either here or on Discord (@Xarvus). Good luck with this and all the projects to come!

MathGame.JsPeanut/MathGame.JsPeanut/Program.cs
```diff
@@ -0,0 +1,234 @@
...
+                CheckResultAndAwardScore(substraction);
+                break;
+            case "M":
+                symbol = "*";
+                CreateAndShowCalc(multiplication, symbol);
+                CheckResultAndAwardScore(multiplication);
+                break;
+            case "D":
+                symbol = "/";
+                createAndShowCalc(division, symbol);
```
edvaudin, line 70:
```suggestion
                CreateAndShowCalc(division, symbol);
```

MathGame.JsPeanut/MathGame.JsPeanut/Program.cs
```diff
@@ -0,0 +1,234 @@
...
+        }
+
+    }
+
+    public static void StartGame()
+    {
+        List<Operation> previousGames = new();
+        Operation sum = new();
+        Operation substraction = new Operation();
+        Operation multiplication = new Operation();
```
edvaudin, line 36:
```suggestion
        Operation substraction = new();
        Operation multiplication = new();
```

MathGame.JsPeanut/MathGame.JsPeanut/Program.cs
```diff
@@ -0,0 +1,234 @@
+class Program
+{
+    
+    public static void Main(string[] args)
+    {
+
+
```
edvaudin, line 7:
```suggestion
```

edvaudin (approved):
Hi @JsPeanut overall much better! A few little things I wanted to tidy up and also sometimes still had some issues with the input flow, but overall still worked well! Good job and good luck in your next projects!

edvaudin (approved):

PR #29, opened 2023-03-15
My first project!

JoshuaMellon:


MathGame.JoshuaMellon/Program.cs
```diff
@@ -0,0 +1,2 @@
+﻿// See https://aka.ms/new-console-template for more information
```
TheCSharpAcademy, line 1: @JoshuaMellon  Looks like your changes didn't come through. Let me know if you have difficulties and we can go through it together.

TheCSharpAcademy (changes requested):

JoshuaMellon:
I've uploaded my changes now, apologies for before!

PR #35, opened 2023-04-03
My version of the MathGame

schnopsi:


Schnopsi.MathGame/ConsoleInteractions.cs
```diff
@@ -0,0 +1,124 @@
...
+			{
+				Console.Write("\nInvalid input!\nPlease enter a valid option: ");
+			}
+		} while (isValidInput == false);
+		return userInput;
+	}
+
+	private static void ShowMainOptions()
+	{
+		Console.WriteLine("Choose one of:\n");
```
TheCSharpAcademy, line 35: Upon choosing an option, the user should be presented with an equation to solve. Then, based on the response, it should increment the score. You can make the number of rounds optional or fixed. But I couldn't see the game element in this version, only the calculations.

TheCSharpAcademy (changes requested):

Schnopsi.MathGame/ConsoleInteractions.cs
schnopsi: Fixed the whole thing to fit the tutorial

PR #36, opened 2023-04-03
MathGame

ItsSt0rm:


MathGame.ItsSt0rm/Program.cs
```diff
@@ -0,0 +1,189 @@
...
+                    if (responseSubstraction == firstNumber - secondNumber)
+                    {
+                        correctAnswers++;
+                        Console.WriteLine("----------- Correct! -----------");
+                    }
+
+                    break;
+
+                case 3:
+                    Console.WriteLine($"{firstNumber} * {secondNumber} = ?");
```
TheCSharpAcademy, line 134: Hi! Your code is looking tidy, but looks like there's a bug. If I input the incorrect game in my first answer, the game is over: 

image

TheCSharpAcademy (changes requested):
The code is looking clean and tidy, could you please just work on that bug?

Also, I suggest you do a little refactoring so your switch statements are leaner. You could call a method for each game in each case instead of having the game's logic in the switch itself. Not mandatory but it would be good practice. 😊

PR #38, opened 2023-04-05
Mathgame Contribution

jwhitt3r:
My attempt at the math game.

### DIFF
--- a/MathGame.jwhitt3r/MathGame.jwhitt3r/Game.cs
+++ b/MathGame.jwhitt3r/MathGame.jwhitt3r/Game.cs
@@ -57,9 +57,17 @@ namespace MathGame.jwhitt3r
         /// <param name="gameScore">The Score object that is used to tracker the users successful answers</param>
         /// <param name="difficulty">The difficulty of the game</param>
         public Game(char coreSymbol, Score gameScore, int difficulty) {
-            this.Symbol = coreSymbol;
+            if (coreSymbol == 'r')
+            {
+                this.Symbol = this.GetRandomSymbolFromList();
+            }
+            else
+            {
+                this.Symbol = coreSymbol;
+            }
             this.results = gameScore;
             this._difficulty = difficulty;
+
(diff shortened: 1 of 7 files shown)


MathGame.jwhitt3r/MathGame.jwhitt3r/Game.cs
```diff
@@ -0,0 +1,225 @@
+﻿using System.Data;
+
```
TheCSharpAcademy, line 2: Is this the expected behaviour? If I reply incorrectly I don't get a new question: 

image

MathGame.jwhitt3r/MathGame.jwhitt3r/Game.cs
```diff
@@ -0,0 +1,225 @@
...
+        }
+
+        /// <summary>
+        /// GenerateEquation builds the structure of the math problem to be presented to the user.
+        /// This function calls the GenerateNumbers function to get the array of integers, constructs
+        /// the default difficulty, which is two numbers and the core symbol and then extends the equation
+        /// to incorporate additional symbols and additional numbers based on the difficulty of the game.
+        /// </summary>
+        /// <returns>A string is returned that presents the equation to the user</returns>
+        private string GenerateEquation()
```
TheCSharpAcademy, line 112: As per requirements, only divisions that result in integers should be allowed. 

image

TheCSharpAcademy (changes requested):
Very impressed with your code. Looks like you have formal training, or have worked in enterprise code bases. 😁. Could you please just work on those two issues raised?

MathGame.jwhitt3r/MathGame.jwhitt3r/Game.cs
TheCSharpAcademy: @jwhitt3r No it's ok, I thought this was unintended but I'm not too fussy I didn't specify the precise wanted behaviour in the requirements

PR #39, opened 2023-04-07
Simon's math game

Simon9566:
Looks to be fine unless I created a bug after the last test :)


Simon9566_Math_game/Program.cs
```diff
@@ -0,0 +1,223 @@
...
+Console.ResetColor();
+
+// Getting the difficulty and constructing main play loops based on them
+
+string difficulty = "tmp";
+var difficulties = new Dictionary<string, string>();
+difficulties.Add("E", "easy");
+difficulties.Add("M", "medium");
+difficulties.Add("H", "hard");
+difficulties.Add("S", "special");
```
TheCSharpAcademy, line 30: At the moment for whatever reason when I play a game, I get only one operation. Example: 
image.

You can either let me choose which operation I want or make it random so it alternates

Simon9566_Math_game/Program.cs
```diff
@@ -0,0 +1,223 @@
...
+    }
+}
+
+
+void Ask_addition()
+{
+    a = random.Next(100, 999);
+    b = random.Next(100, 999);
+    Console.WriteLine($"What is {a} + {b}?");
+    guess = Convert.ToInt32(Console.ReadLine());
```
TheCSharpAcademy, line 73: At the moment there's no validation so if I input something that can't be parsed I get an error. You can create a loop that keeps running while the attempt is invalid. Let me know if you need help. 

image

TheCSharpAcademy (changes requested):
Hey there! It's looking great! Good work so far! Please have a look at the two comments I made in the review and also make sure the program only presents to the user divisions that result in integers, as per the project's requirements. 

Oh, also as per requirements I need to have an option to see a list of previous games played. Let me know if you need help! You can also see how it's done in this course: https://www.youtube.com/watch?v=iLByqqLGYEc&lc=Ugzhol3KJJuXmjd86Id4AaABAg

image

TheCSharpAcademy (changes requested):
Hey! You were doing well! But now you transformed the app in a simple calculator 😟

This is the behavior I expect: Upon choosing an operation I get given questions that I have to answer. You can let me choose how many questions or have a fixed number (up to you). 

The history can consist of date, game type and number of correct answers. Let me know if you need help!

Simon9566:
Wait it changed THIS pull request?
I wanted to open a new one with the calculator app I guess I messed up
somewhere with pushing the changes, I'll try to fix it later

TheCSharpAcademy:
@Simon9566 hey there! Just stopping by to see if you need any help. 😁

TheCSharpAcademy (changes requested):
Hi there! At the moment your project doesn't compile. There are several errors, such as void methods trying to return something. 

image

Simon9566:
Visual studio refused to push the commit that fixes those and i went to
sleep due to my time zone :)

TheCSharpAcademy (changes requested):
Good stuff! You're close! I'm usually not super strict with the first project except for the requirements. The users should only be presented with divisions that RESULT IN AN INTEGER, as per requirement 2. 

image

Simon9566_Math_game/Program.cs
```diff
@@ -160,18 +161,19 @@ void Ask_division()
 {
     a = Get_large();
     b = Get_small();
-    result = $"{a / b}";
+    int div_result = a / b;
+    result = $"{div_result}";
     Console.WriteLine($"What is {a} / {b}?");
     Validate_guess();
-    result = $"{a} / {b} = {a / b}";
```
TheCSharpAcademy: Hey! I'm sorry, I think you still didn't understand. I know it's a bit frustrating, and you just want to finish the project, but I cant let this requirement pass. So let me explain again: At the moment your program presents to the users equations that result in non-integers. Example: 
image. 

This shouldn't happen. This equation should never be shown to the users. Another example: 9/4 shouldn't be allowed, 9/3 should. You can use a loop to keep generating numbers until you get a suitable equation. Let me know if that makes sense?

TheCSharpAcademy (changes requested):

Simon9566_Math_game/Program.cs
Simon9566: Ah so it should only roll divisions that have a result a % b == 0. You are absolutely right I did misunderstand that part :)

PR #49, opened 2023-04-30
Pushing final project files from separate repository

csm-stough:
I worked on this in a separate repo before reading the review article, so this is only my final project state. Going forward, I will first clone the review repo :D


csm-stough:
Hello, I accidentally submitted the incorrect link in the "Submit Project" link from the C# Academy site for this project. I'm not sure if that submission is just for keeping track of user progress or not, but this PR has the correct project files, Thanks!

PR #54, opened 2023-05-20
Final Changes

KdRome:


TheCSharpAcademy (changes requested):
Hi there! Thanks for submitting your project for review, well done! There are a couple of things I'd like you to fix though. First, there's a bug when I choose to view the games history, as per picture:

image

Also, most of the comments in your code are unnecessary. You should litter the code with comments where the code is self-explanatory. Leave comments only where the functioning of your app absolutely can't be inferred from the code. Examples of excessive comments: 
image

KdRome:
Thank you for the feedback!

PR #58, opened 2023-05-28
Submission for review

vincentyarmstrong:


MathGame.VincentyArmstrong/MathGame.VincentyArmstrong/Game.cs
```diff
@@ -0,0 +1,80 @@
...
+                result = firstNumber * secondNumber;
+                Console.WriteLine($"{firstNumber} x {secondNumber}");
+            }
+            if (Type.Equals(GameType.Division))
+            {
+                int[] divisors = Utilities.GetDevisors(firstNumber);
+                secondNumber = divisors[random.Next(divisors.Length)];
+                result = firstNumber / secondNumber;
+                Console.WriteLine($"{firstNumber} ÷ {secondNumber}");
+            }
```
edvaudin, line 63: These should really be handled in a separate method for getting the result and printing the question.

MathGame.VincentyArmstrong/MathGame.VincentyArmstrong/Game.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            bool parsed;
+            int userResult;
+            do
+            {
+                parsed = int.TryParse(Console.ReadLine(), out userResult);
+                if (!parsed) Console.WriteLine("Please enter a number");
+            } while (!parsed);
+            if (userResult == result) Score++;
```
edvaudin, line 72: This process could be extracted into another method for handling the quiz input loop.

MathGame.VincentyArmstrong/MathGame.VincentyArmstrong/UserInterface.cs
```diff
@@ -0,0 +1,125 @@
...
+            default:
+                break;
+        }
+    }
+
+    private void Setup(GameType gameType)
+    {
+        Questions();
+        Difficulty();
+        Game game = new Game(gameType, questions, hardMode);
```
edvaudin, line 66: You can technically remove the second `Game` from this initialisation as the first one defines it.
```suggestion
        Game game = new(gameType, questions, hardMode);
```

MathGame.VincentyArmstrong/MathGame.VincentyArmstrong/UserInterface.cs
```diff
@@ -0,0 +1,125 @@
...
+        games.Add(game);
+    }
+
+    private void Questions()
```
edvaudin, line 70: This method name is unclear. It does not provide enough information as to what is going on in the method.

Consider:
```suggestion
    private void GetNumberOfQuestionsFromUser()
```

MathGame.VincentyArmstrong/MathGame.VincentyArmstrong/UserInterface.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        var input = Console.ReadLine();
+        while (!int.TryParse(input, out questions) || questions <= 0)
+        {
+            Console.WriteLine("Please enter a whole number larger than zero");
+            input = Console.ReadLine();
+        }
+    }
+
+    private void Difficulty()
```
edvaudin, line 83: Same as the above method, I would suggest a more detailed name so that it is clear what the method does.

```suggestion
    private void GetDifficultyFromUser()
```

MathGame.VincentyArmstrong/MathGame.VincentyArmstrong/UserInterface.cs
```diff
@@ -0,0 +1,125 @@
...
+            }
+            if (difficulty == "E") break;
+            Console.WriteLine("Please select a difficulty");
+        }
+    }
+
+    private void ViewHistory()
+    {
+        Console.Clear();
+        if (games.Count == 0) Console.WriteLine("No games have been played");
```
edvaudin, line 106: If there are no games, then you could `return` out of this method at that point without having to run the rest.

edvaudin (changes requested):
Well done on completing your Math Game! This was a very good submission. My suggestions primarily focus on extracting methods where possible and better naming of methods. If you have any questions, I am always available on Discord (Xarvus).

edvaudin (approved):
Very clean set of changes here! Good job!

PR #66, opened 2023-06-17
project completed

CosmicPhoenix2029:


MathGame.CosmicPhoenix2029/GameItem.cs
```diff
@@ -0,0 +1,154 @@
...
+                break;
+            case "S":
+                Question = $"What is {NumOne} - {NumTwo} - {NumThree} - {NumFour}";
+                CorrectAnswer = NumOne - NumTwo - NumThree - NumFour;
+                break;
+            case "M":
+                Question = $"What is {NumOne} * {NumTwo} * {NumThree} * {NumFour}";
+                CorrectAnswer = NumOne * NumTwo * NumThree * NumFour;
+                break;
+            case "D":
```
TheCSharpAcademy, line 126: Divisions that result in non-integers shouldn't be presented to the user.

TheCSharpAcademy (commented):
Hi there! Good work! The project looks great. I'll just ask you to review your division logic, as per Requirement 2: 

image

MathGame.CosmicPhoenix2029/GameItem.cs
```diff
@@ -124,7 +127,7 @@ string HardQuestion()
                 CorrectAnswer = NumOne * NumTwo * NumThree * NumFour;
                 break;
             case "D":
-                Question = $"What is {NumOne} / {NumTwo} / {NumThree} / {NumFour}";
+                Question = $"Rounded down to the nearest whole number, what is {NumOne} / {NumTwo} / {NumThree} / {NumFour}";
```
TheCSharpAcademy, line 130: Just to clarify, your program should never show this equation to the user, since it doesn't result in a whole number. Every division presented to the user needs to result in an integer. Let me know if that makes sense?
image

TheCSharpAcademy (changes requested):

CosmicPhoenix2029:
Hi, many thanks for the review. I believe I understand the requirement now. I will make the requested changes. 
As I want this to be unique, I want to keep the logic for having an increased number per difficulty. This makes it so that the logic for division is not process efficient at all, however believe for uniqueness, it is worth keeping in. 

In order to improve the process efficiency when the division case is matched, I have tried to min max the random numbers to give it the best chance of being a whole number, without having to iterate, yet I'm no mathematician and so am aware there will always be room to improve the efficiency.

TheCSharpAcademy:
@CosmicPhoenix2029 No problem, as long as questions presented to the user can only result in integers. 😊

CosmicPhoenix2029:
Hi - I have amended the division logic so instead of increasing the number of numbers for each difficulty, the 2nd divisor will be in a specific range per difficulty. For Easy mode, the divisor is between 1 and 5, Medium mode is between 6 and 15, and hard is between 16 and 100. I did try to work with the logic previously implemented, however was running into too many issues. Really appreciate your reviews on this project, it helps me a lot.

TheCSharpAcademy:
@CosmicPhoenix2029 Well done! I'm approving the project. I'm not overly strict with comments in the first couple of projects, but for the future, please have a look at Item 3 in this article: https://thecsharpacademy.com/article/58. 

It's ok to comment on code when studying, but when pushing it, make sure you comment only when absolutely necessary 😊. Code littered with redundant comments is heavily frowned upon. 

image

PR #68, opened 2023-06-20
re-adding project

kmakai:
file was not showing up in repo


kmakai:
some reason committing a folder I don't have write access to do a new one for a re pull

MathGame.kmakai/MathGame.kmakai/Calculator.cs
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    public static int Multiply(int a, int b)
+    {
+        return a * b;
+    }
+
+    public static int Divide(int a, int b)
+    {
+        return a / b;
```
TheCSharpAcademy, line 22: image

According to the second requirement, your program shouldn't present to the user equations that don't result in an integer. This shouldn't be showed to the user: 

image

TheCSharpAcademy (commented):
Well done on finishing the project! Could you please just fix requirement 2 as per comment? 😁

PR #71, opened 2023-06-24
Submit math console project for review

marvinobig:
# My math console app

## Topics learned building this app
- Variable data types
- Getting user input
- Working with string data
- Creating and accessing lists
- List methods
- Branch logic
- Catching errors using the try...catch statement
- String methods
- User-defined methods
- Passing parameters into methods
- Debugging in Visual Studio
- do...while loops
- for loops
- Lists vs arrays
- for...each loop
- Classes & objects
- Access modifiers
- Namespaces
- Static methods
- Fields
- Publishing Code

### DIFF
--- a/Math.marvinobig/App.cs
+++ b/Math.marvinobig/App.cs
@@ -1,0 +1,105 @@
+﻿namespace Math
+{
+    internal class App
+    {
+        private readonly string[] availableChoices = new string[5] { "a", "s", "m", "d", "q" };
+        private readonly int[] mathNumsList = { 23, 45, 56, 78, 23, 78, 46, 98, 2, 10 };
+        private List<string[]> history = new();
+        protected string? username;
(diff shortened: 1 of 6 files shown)


TheCSharpAcademy (changes requested):
@marvinobig Well done on completing your first project! Here are a few things that need to be addressed:
1 - Once I play a game the program terminates. Instead the user should go back to the main menu.

2 - Divisions that result in non-integers shouldn't be presented to the user, as per requirement 2.
image
image

3 - As per requirement four, you there should be an option for users to see previous games. You can use List collections for that. 
image

Let us know if you have questions!

TheCSharpAcademy:
@marvinobig Let me know if you need help! 😊

TheCSharpAcademy:
@marvinobig feel free to reopen the PR if you want to proceed 😁

PR #80, opened 2023-07-25
Finished project w0lvesvv

w0lvesvvv:
Be as brutal as you can be with the review :)

### DIFF
--- a/MathGame.w0lvesvvv/MathGame.w0lvesvvv/Game.cs
+++ b/MathGame.w0lvesvvv/MathGame.w0lvesvvv/Game.cs
@@ -1,0 +1,284 @@
+﻿namespace MathGame.w0lvesvvv
+{
+    public class Game
+    {
+        // You should record previous games in a List and there should be an option in the menu for the user to visualize a history of previous games.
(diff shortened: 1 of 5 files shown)


MathGame.w0lvesvvv/MathGame.w0lvesvvv/Game.cs
```diff
@@ -0,0 +1,284 @@
...
+        {
+            get
+            {
+                switch (option)
+                {
+                    case "A":
+                        return "+";
+                    case "S":
+                        return "-";
+                    case "M":
```
TheCSharpAcademy, line 30: @w0lvesvvv here you could make it case insensitive for better user experience. 
image

TheCSharpAcademy (changes requested):
Usually I'm not too fussed but since you asked me to be brutal, it would be better to show more info in the results. On top of what you already have you could add the date and the score/number of questions.

But well done! The only requirement you didn't follow was the division issue. I can't merge without that solved. 😔

TheCSharpAcademy:
@w0lvesvvv Well done! I really like the colorful console 😁

PR #84, opened 2023-08-04
Finished Project

lordWalnuts:
Added all the functionalities. Took me a while but learned a lot especially about Git Workflow


TheCSharpAcademy:
@lordWalnuts Well done on your first project! I'm approving it now. For the future, keep an eye on redundant comments. Code should be self-explanatory and some of your comments are saying things that can be inferred by the code itself. That's why you should name your variables and methods in an informative way. 😁 
image

PR #89, opened 2023-08-21
Project done.

VocalNight:


TheCSharpAcademy (changes requested):
@VocalNight thanks for making the changes! I don't think you're resetting the operands in the division. Could you please check that.

VocalNight:
Wooops. Indeed i wasn't, fixed now.

PR #90, opened 2023-08-22
[HappyReni] MathGame submit

HappyReni:


MathGame.HappyReni/GameManager.cs
```diff
@@ -0,0 +1,136 @@
...
+
+                foreach (var h in _history)
+                {
+                    Console.WriteLine($"{h.Time,-25}{h.Question,-20}{h.CorrectAnswer,-20}{h.Answer,-20}{h.Result,-5}");
+                }
+            }
+            Console.WriteLine("".PadRight(100, '='));
+        }
+
+        private (bool res, string str, int val) GetInput(string s)
```
TheCSharpAcademy, line 121: In general I recommend you name your parameters with a word that means something. All of your parameters are named s, which means the readers of your code will have to spend some time figuring out what's going on. Ideally code should read like a book. 😊

MathGame.HappyReni/History.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.Linq;
```
TheCSharpAcademy, line 2: Please remove unused usings as per our conventions: https://thecsharpacademy.com/article/58

TheCSharpAcademy (changes requested):
@HappyReni Hey there! Thanks for submitting! Well done. The project is looking great.

There are a couple of weird behaviors though. It seems like the score doesn't get reset once a game is over. After playing a few times I got this after playing a 3-question game:  
image

Also, when asked the number of questions, I typed a random string (lala) and it was announced the game was over and the score. COuld you please fix these issues?

PR #93, opened 2023-08-27
Finished project

maccer989:


CarDioLogic:
Good job! It is working according to all the requirements.

PR #95, opened 2023-08-28
Strawfervor's solution

strawfervor:


TheCSharpAcademy:
@strawfervor Well done on completing the project! For future reference, code reviews will get more strict on the following: 
- Leaving unnecessary comments
- Method names should be pascal case
- Requirements have to strictly be met (in this project division operands should go from 0 to 100)
- Remove unused usings
- Not a requirement but you can use the new namespaces convention for better use of space.

You can find more about our code conventions here: https://thecsharpacademy.com/article/58

PR #97, opened 2023-09-01
project finished

Aapje185:


MathGame.Aapje185/UI/UserInterface.cs
```diff
@@ -0,0 +1,52 @@
...
+
+    public void GetOptionFromUser()
+    {
+        int userChoice;
+        while (!TryParse(Console.ReadLine(), out userChoice) || !_options.ContainsKey(userChoice))
+        {
+            Console.WriteLine($"The chosen option is not available, please enter a valid option.");
+        }
+
+        if (userChoice == 5) return;
```
TheCSharpAcademy, line 46: I'm getting this error when I choose division 

image

TheCSharpAcademy (changes requested):
@Aapje185 Thanks for submitting! It's looking great. Could you please just fix the issue in the comment? 😊

MathGame.Aapje185/UI/UserInterface.cs
Aapje185: Thank you for reviewing. Bug should be fixed now.

TheCSharpAcademy (changes requested):
@Aapje185 almost there! As per second requirement, you shouldn't show equations that result in non-integers: 
image

image

TheCSharpAcademy:
@Aapje185 I'm closing the PR as it's been 2 weeks, but feel free to reopen it if you want to continue 😊

PR #122, opened 2023-10-05
Add complete project

dnym:
Full documentation at [my personal repository](https://github.com/dnym/tcsa01).

### DIFF
--- a/MathGame.dnym/MathGame.dnym.sln
+++ b/MathGame.dnym/MathGame.dnym.sln
@@ -1,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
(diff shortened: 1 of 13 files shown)


TheCSharpAcademy:
@dnym Well done, your project looks great! I'm approving it now. Just a few suggestions for the future:
1 - This is not very readable. I would rearrange it or find another solution that would be easier to maintain. I'd have to spend several minutes trying to understand what's going on here in case I had to work with this code. I recommend you make your code more verbose and less cryptical so it's easily readable.
image

2. I would say the same about this section. It's more organized but the field names aren't self explanatory. Range of what? Domain of what? 
image

3. This isn't a very good practice. Exceptions are expensive and if you're going to catch them you should do something with them. 
image

Other than that, really good project 😊

PR #133, opened 2023-10-17
Program.cs

SadePauw:


TheCSharpAcademy:
@SadePauw Well done! You got your Green Belt! 😁
Just a few considerations: 
1 - It would be good to clear the console after a game, for better readability
2 - It would be a nice addition to have a message when the users submit invalid input. (e.g. an option that doesn't exist or a string when an integer is expected.

PR #137, opened 2023-11-06
project added

szejku11:


TheCSharpAcademy:
@szejku11 well done! The project looks great. The only thing I'd say is that I chose a difficult game and got easy multiplications. But I'm approving since this isn't a requirement.

image

PR #138, opened 2023-11-07
Project finished!

MicVerg:
I followed the instructions through youtube first, then remade it afterwards.

I did have to check the code of the first version again, for usage of the gamemodel and how to use the list correctly.


TheCSharpAcademy:
@MicVerg Well done! The code is neat and tidy! 😎

PR #139, opened 2023-11-09
Finally project Finished

idrisKurtar:


TheCSharpAcademy:
@idrisKurtar Well done!  I'm approving the project now. You're now a Green Belt!!

A few things for the future (as I'll get more strict)

1 - Remove any unused code and unnecessary empty lines. Example:

image 

2 - Make your indentation pristine. This one, for example, is wonky:

image

3 - This wasn't a requirement (it will be in future projects), but right now If I input a word instead of a number I get an error. It would be good to handle that:

image

4 - I recommend you upgrade your .NET to at least version 6, there are significant differences from .NET framework.

image

idrisKurtar:
Thank you so much Pablo. I am very grateful

PR #140, opened 2023-11-11
Completed MathGame project and uploading for code review

nariks:


TheCSharpAcademy:
@nariks Approving the project. As we've discussed, I would avoid nested loops (specially 3 of them) when possible as it makes it difficult to read/debug. 😊

PR #142, opened 2023-11-11
Complete MathGame project

Lu1zLuna:
Uploading for code review


TheCSharpAcademy:
@Lu1zLuna  Well done! Project approved! Well done on completing the challenges 😁
Just a couple of things:

You could use an enum for the difficulty as well. But even if you don't, refrain from commenting unless absolutely necessary. Code should be self-explanatory. If you really want to comment, it's better to write [summaries.](url) 

image

Also, you don't need to invoke Helpers since the method is being called from within itself. 
image

PR #147, opened 2023-11-21
Added Game difficulty 

nwdorian:


TheCSharpAcademy:
@nwdorian Well done! 🤓 It's great that you completed the challenges. Solving those problems will definitely make you a better developer.

Just one thing I would add is, in the menu, when I try to input an option that's not valid, I can see you have a default message, but the user can't see it because the loop immediately runs again. So you could add a "press any key to continue" type of command so the user has the opportunity to see the message.

PR #148, opened 2023-11-22
Project Finished!

GustavoBarrosGit:
Did all the challenges, the only thing is that division has no difficulty, the others have.


TheCSharpAcademy:
@GustavoBarrosGit Well done! 😁It's very good you completed the challenges. These are the types of exercises that will make you a better developer.

One thing for the future. You can improve readability by removing one level of indentation. You can do that by putting a semi-colon in front of the namespace: 
image

PR #149, opened 2023-11-27
Project finished! (?) 

joao-luizi:
I attempted to implement the State Pattern for this solution.
Design Patterns are a bit of a challenge for me considering my current knowledge base - so i figured i would tackle them. 

Got a little confused with MVC because every State handles presentation and user input (so i guess it would be the View in MVC) and the Controller in MVC that i'm guessing could be maybe the Context class ?!
Also i didn't know where to put the State abstract class. The closest i could find in MVC would be Models. I ended putting it in a Abstractions folder.

I had a great fun completing this project - andit was good practice.


TheCSharpAcademy:
@joao-luizi Very interesting project! I wasn't aware of this design pattern. I'm glad the project was useful. Well done! 🤓🤓

PR #152, opened 2023-12-12
project finished

ndthuy112:


TheCSharpAcademy:
@ndthuy112 Well done! Your code is neat and tidy 🤓 I think the only thing I would have done differently was

image

PR #154, opened 2023-12-15
Finished Math Game

Doc415:
Finished math game with all challenges complete
There is also a boot screen

### DIFF
--- a/Program.cs
+++ b/Program.cs
@@ -1,0 +1,531 @@
+﻿using System;
+using System.ComponentModel.Design;
+using System.Diagnostics;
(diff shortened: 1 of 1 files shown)


TheCSharpAcademy:
@Doc415 Well done! Your project is approved 😁
For future reference, you will need to submit pull requests inside a folder with this name: ProjectName.YourGithubUserName. The folder needs to contain the code plus the .csproj (and .sln if applicable).

Also, it would be a nice challenge to have all operations done in one single method, since there's a lot of repeated code in your solution. 

Looking forward to seeing your next project. 🤓

PR #156, opened 2023-12-15
Finished MathGame

b4lkaan:


TheCSharpAcademy:
@b4lkaan Did you submit the project in the Academy's platform? I can't see it in the database. 🥺

TheCSharpAcademy:
@b4lkaan Well done! Project approved! Just a couple of considerations:

- Be careful with a loop that calls the method that contains it. It can get out of control and difficult to debug very quickly. It would be better to have a boolean that will be switched so the loops runs again.
image

- Don't leave unnecessary comments. Code should be self explanatory. Only comment when there's something in the code that requires clarification.
image

PR #162, opened 2023-12-26
Finished the project.

eddyfadeev:


TheCSharpAcademy:
@eddyfadeev Very nice project! I really like the way you organized the menu and created summaries for methods. Well done.

For future reference, try not to leave unnecessary comments. Code should be self-explanatory. It's ok to add comments as part of your study in your own repository but when pushing code leave them out. Here's an example of comments that I thought were unnecessary:

image

This isn't a requirement, but you could have created a custom type to encompass all these arguments. 
image

Well done!

eddyfadeev:
Thank you for your review and suggestions!

PR #166, opened 2024-01-09
Add files via upload

TacticATom:


TheCSharpAcademy:
@TacticATom Well done! I'm approving the project 🤓
For future reference, I'll get more strict on these: 
1 - Don't leave unnecessary comments or unused code:
image

2 - Your project name should be: ProjectName.GithubUsername

3 - For better readability, try to separate your code into related chunks. At the moment the whole thing is just one big block:
image

4 - It would be good to actually track the user's score, as in: out of x games, how many answers were correct? 

Looking forward to your next projects 😁

PR #168, opened 2024-01-12
Code review

SecondNewtonLaw:
I felt like just commenting the code, just for comedic effect, most of the code can be understood by just looking at it, still that won't help out with edge cases for future maintainers, so while most comments are for comedic effect, others aren't, so yeah

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -475,3 +475,10 @@ $RECYCLE.BIN/
 
 # Windows shortcuts
 *.lnk
+
+# Rider git ignore.
+bin/
+obj/
+/packages/
+riderModule.iml
+/_ReSharper.Caches/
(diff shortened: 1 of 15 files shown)


SecondNewtonLaw:
Static analysis is wrong, you gotta have the System.Numerics (Chose to use BigInteger, mostly because if numbers got over Int64 (Not really a problem being honest with the number range this is going for, but just in case)), so ignore that.

TheCSharpAcademy:
@SecondNewtonLaw Well done! I assume you've got previous C# experience? 
Not many people use interfaces this early on. And I don't think I've seen anyone use BigInteger in any project 😁

SecondNewtonLaw:
> @SecondNewtonLaw Well done! I assume you've got previous C# experience? Not many people use interfaces this early on. And I don't think I've seen anyone use BigInteger in any project 😁

You can say so, 4 years of C# makes things happen, mostly taking it up to have some fun

PR #169, opened 2024-01-19
willN123: 1st submission for Math Game project.

willN123:


MathGame.willN123/Program.cs
```diff
@@ -0,0 +1,430 @@
...
+            else
+            {
+                Console.WriteLine("Invalid input. Please enter a number.");
+            }
+        } while (!validInput);
+
+        Console.Clear();
+
+    } while (gameRunning);
+}
```
edvaudin, line 196: This method should be split into several smaller methods since it has multiple side effects beyond displaying the Game Screen. For example it includes all the logic for selecting a random operation.

MathGame.willN123/Program.cs
```diff
@@ -0,0 +1,430 @@
...
+void SaveScore(string operation)
+{
+    if (gamesPlayed < 9)
+    {
+        gameResults[gamesPlayed] = $" {gamesPlayed + 1} |  {operation}   |  {score}/{questionCount}  |  {timeTaken}";
+    }
+    else if (gamesPlayed >= 9)
+    {
+        gameResults[gamesPlayed] = $" {gamesPlayed + 1}|  {operation}   |  {score}/{questionCount}  |  {timeTaken}";
+    }
```
edvaudin, line 248: This took a while for me to understand the purpose of this method. Can you find a way to refactor this to explain its purpose through code?

edvaudin (changes requested):
Overall this is a good project with clean UI and codebase. I would suggest however that some of the methods are quite large in size and may do multiple things such as displaying options and processing them, when those are two separate behaviours. If you could particularly look at separating out the `GameScreen()` method, that will improve the readability of your code. Any questions, please let me know :)

edvaudin (approved):
The changes look good to me! Well done!

edvaudin (changes requested):
Sorry I didn't see this, but before this is approved, you need to fix the code analysis issues that have come up. The methods GetUserAnswer() and SetAnswerText() return values but those values are never used. I think you are simply setting global variables, in which case these methods should return void. Can you fix that and then we can get this project finished :)

willN123:
Hi, sorry the issues should be fixed now and it shows that the code checks pass. Thanks for taking your time to do these reviews :)

TheCSharpAcademy:
@edvaudin is this good to go? 😁 Thanks for reviewing

PR #170, opened 2024-01-22
Project completed!

WinnieNgina:
Be brutal as much as you can. I need to take my coding skills to the next level!


TheCSharpAcademy:
@WinnieNgina  Project approved! I don't have much to say in regards to the code itself, just a few tidy-up things. Always think of cleaning up after you're done. Keep your code always PRISTINE!

1: Get rid of unnecessary comments: 
image

2: Remove one level of indentation using a semi-colon after a namespace: 
image

3: Add a bit more separation between related code chunks, or it will quickly turn into spaghetti. I've changed this method as an example: 
image

4: Remove unnecessary empty lines: 
image

Other than that, all good! Looking forward to seeing your next project!

PR #178, opened 2024-02-05
Complete Math Game project

Dejmenek:


TheCSharpAcademy:
@Dejmenek Superb work! Well done in completing the challenges. 🤓🤓

A couple of things to point out: 
You could use a default case in your switch statement to handle an incorrect input. 
image

In regards to organization, try to abide to the separation of concerns principle and only put relevant methods in a given class. As an example, a game class should have methods concerning a game only. In this case it has methods related to user interface and validation. You could outsource them to an UserInterface and a Validation classes: 
image

Keep up the good work! 😁

PR #183, opened 2024-02-19
finished project

Deviationlark:
did most of the project on my own after doing the c# course on freecodecamp, only added the classes,objects and lists from the yt tutorial to make it cleaner


VocalNight:
Hey man, great job! Everything is working properly. 
The only thing i could find that you should take a look is that when selecting a difficulty, i can input anything other the prompt numbers and it will still work instead of giving me an error. This will mess up the history list.

If you fix that and the codacity analysis, this will be a 10/10 project.

image

Deviationlark:
Fixed it

VocalNight:
All good!
Well done.

PR #186, opened 2024-02-22
MathGame Project

Rxxyxd:
If anything can be simplified please let me know:)

Sorry bout the comments.


TheCSharpAcademy:
@Rxxyxd Hey there! Well done, project approved. 😊

A couple of things I should mention: 
1- It's ok to have comments in your own repository for studying purposes, but you'll be asked to remove these comments from now on. This is because this is a heavily frowned upon habit in the industry. Code should be self-explanatory. Only leave comments when absolutely necessary to clarify something that can't be inferred by the code.

2- For organization, it would be better to call a method for each operation, or even better, one method that will handle all operations with the same code, only changing a small amount of variables. Your switch statement ended up very bloated. 

3 - For readability, split your code in chunks, otherwise it will become spaghetti very quickly. For example, I've split your division case into related chunks: 
image

4 - Throw exceptions when ... exceptions happen. Validation isn't an exception, it's a set of expected errors that can be handled with a loop. (e.g. while (!int.TryParse) { " Error, try again " // get input again }

Looking forward to your next projects! 🤓

PR #190, opened 2024-02-23
Fixed submodule error

kronyer:


Doc415 (changes requested):
Hey there. You code looks good. There is a part that code throws exception.
When I choose whatever in division operation and enter characters instead of numbers
code throws exception. Can you please add a validation to division whatever part?

image

PR #195, opened 2024-02-27
Initial commit

obitom67:
Initial pull request for the Math Game project


VocalNight (approved):
Good job! Everything is working fine so i'll be approving the project.
I would like to point out one thing though that you will need for your next console projects.

image

The console can get a bit full if you use the application a lot, or have a lot of operations back to back. See if you could find a way to clean the interface a bit, (You can use Console.Clear() for this!)

PR #199, opened 2024-03-05
My finished project files

HopelessCoding:


VocalNight (approved):
Good job on this project. I specially liked what you did when the user inputs something wrong while selecting the difficulty. Defaulting to easy is a nice touch.

One thing i will mention, if you use the program too much, the UI get's clogged.

image

Think of ways you can solve that in the next projects.
Well done!

PR #202, opened 2024-03-12
Review Submission

Joe-Whitehead:
Please Review my submission

### DIFF
--- a/JoeWhitehead.MathGame/Helpers.cs
+++ b/JoeWhitehead.MathGame/Helpers.cs
@@ -1,0 +1,172 @@
+﻿using ConsoleMathsGame.Models;
+
+namespace ConsoleMathsGame
+{
+	internal class Helpers
+	{
+		internal static List<Game> games = new();
+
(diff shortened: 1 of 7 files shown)


VocalNight (approved):
Very well done.

I liked the use of enums and how you were able to make the game engine class flexible enough so that all operations can be processed in one method.

The treatment of exceptions and incorrect inputs was also on point.

Very good job not only on doing the main project but also completing all challenges!

JoeWhitehead.MathGame/Helpers.cs
```diff
@@ -0,0 +1,172 @@
...
+			
+			switch (difficulty)
+			{
+				case Difficulty.Easy:
+					low = 1;
+					high = 10;
+				break;
+
+				case Difficulty.Medium:
+					low = 0;
```
VocalNight, line 25: Just one thing you should probably double check, you left the low variable in the medium difficulty have a chance to be 0, which will cause problems. This is just a minor bug, so you can fix on your own time.

PR #203, opened 2024-03-13
Submission MathGame

antonderegt:
Thanks for the review! Any feedback is welcome!

### DIFF
--- a/MathGame.antonderegt/Program.cs
+++ b/MathGame.antonderegt/Program.cs
@@ -1,0 +1,97 @@
+﻿using MathGame.Antonderegt;
+
+List<Result> results = [];
+
+void ShowMenu()
+{
+    Console.Clear();
+    Console.WriteLine("1. Add");
+    Console.WriteLine("2. Substract");
+    Console.WriteLine("3. Multiply");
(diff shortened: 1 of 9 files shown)


VocalNight (approved):
Well done! I really liked how you used an interface to be able to handle all 4 matematical operations.
Code is also well organized.

Good job! I will be approving it.

MathGame.antonderegt/Program.cs
```diff
@@ -0,0 +1,97 @@
...
+    Console.WriteLine("6. Exit");
+
+    Console.Write("\nSelect option: ");
+}
+
+void ShowScores()
+{
+    Console.Clear();
+    Console.WriteLine("Your results");
+    foreach (Result r in results)
```
VocalNight, line 22: The only feedback i feel needed here is to never use single letter variables, always use a good and descriptive name, no matter how obvious their use is.

Someday you will do it out of a habit in a bigger project, come back one week later and scratch your head for a good while.

antonderegt: @VocalNight thank you, I will definitely keep that in mind!

PR #204, opened 2024-03-16
my attempt at math game

Frolikss:


stevietv:
Hi @Frolikss - thanks for your submission, welcome to The C# Academy! Its a nice and well rounded math game that also meets all the challenges, well done!

A few things/pointers I would give to make the game cleaner:

* consider clearing the console more often so that the current menu is the only item displayed.
* consider allowing the user to get an answer incorrect - right now you get stuck in the game until the answer is correct - you could instead track a users Score in each game.
* next time, please also remember to submit your `.sln` file too, in this simple program its not needed but going forward it will be, same with using a namespace.
* choose either `var` or `int` and stick with it :)


keep up the nice work and looking forward to your next submission!

PR #207, opened 2024-03-22
MathGame.JaegerByte

JaegerByte:
My version of the MathGame. Any tips or feedback would be greatly appreciated.
Thank you for the opportunity!


VocalNight (approved):
Well done on your project and for doing some of the challenges.
Overall your code is good! The only thing i would advise you is to think how you could separate your code in differente files/classes instead of writing all of it in a single place.

Otherwise, good job! I will be approving it.

PR #213, opened 2024-03-27
Project by Mothnue

Gincaza:
Hello! That's my first time doing a PR, thank you so much for the tutorial!!

### DIFF
--- a/Math_Game.Mothnue/Program.cs
+++ b/Math_Game.Mothnue/Program.cs
@@ -12,117 +12,114 @@ using System.Timers;
 
 class Program
 {
-    //variavel Random para gerar os numeros
+    // Random variable to generate numbers
     static Random random = new Random();
     static List<string> historyArray = new List<string>();
+
     static void Main()
     {
         while (true)
         {
-            Console.WriteLine("Bem vindo ao jogo de matemática. Meu primeiro projeto em C# btw.");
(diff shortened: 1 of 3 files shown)


Math_Game.Mothnue/Program.cs
```diff
@@ -0,0 +1,308 @@
...
+        MenuDificuldade();
+        string input = Console.ReadLine();
+
+        int diff = ObterDificuldade(input);
+
+
+        for (int i = 0; i < 10; i++)
+        {
+            int x = random.Next(1, diff);
+            int y = random.Next(1, diff);
```
VocalNight, line 226: Hey, good job on the project, everything looks good.

There's just one problem that you would need to fix, since it's a requirement.

Currently your division function is allowing any number to go through, which results in questions like these
image

The division question in specific needs to generate only questions that can be answered by inputing an integer. We can't present to the user questions that would result in decimal number.

Could you you double check this part?

VocalNight (changes requested):

Math_Game.Mothnue/Program.cs
Gincaza: roger that

VocalNight (approved):
Everything good now!
Well done and good luck on your next projects!

PR #214, opened 2024-04-02
Finished project with one of the challenges

jrazorx:


MathGame.jrazorx/MathGame.jrazorx/Program.cs
```diff
@@ -0,0 +1,165 @@
...
+string GameHistoryFormatLine(int gameOccurence, Game currentGame)
+{
+    string gamesHistoryLine = "";
+
+    gamesHistoryLine += gameOccurence.ToString().PadRight(3) + " | ";
+    gamesHistoryLine += currentGame.Mode.PadRight(14) + " | ";
+    gamesHistoryLine += currentGame.Operation.PadRight(9) + " | ";
+    gamesHistoryLine += currentGame.PlayerAnswer.ToString().PadLeft(10) + " | ";
+    gamesHistoryLine += (currentGame.IsWin ? "WIN" : "LOSE").PadRight(6) + " | ";
+    gamesHistoryLine += (currentGame.TimeTakenToAnswer.TotalMilliseconds < 1000 ? "<" : ">") + " 1 second";
```
antonderegt, line 147: Take a look at the StringBuilder class. That class is more efficient when doing string concatenation. 

Your code is totally fine and works, so this is just a suggestion, but in the future when you see string concatenation, think StringBuilder.

MathGame.jrazorx/MathGame.jrazorx/Game.cs
```diff
@@ -0,0 +1,157 @@
...
+        // result of the game : True = WIN, False = LOSE
+        public bool IsWin { get; private set; }
+
+        public TimeSpan TimeTakenToAnswer { get; private set; }
+
+
+        // Constructor, requires game mode letter input (A, S, M, D)
+        public Game(char gameLetter)
+        {
+            Random random = new ();
```
antonderegt, line 32: The `Random` class is an interesting one, because it's output is not really random. Always try to avoid instantiating multiple versions of the `Random` class. You can read more about it [here](https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-random#avoid-multiple-instantiations). 

Your code works, so again this is just a suggestion, but take a look at making `random` a static member of the `Game` class. That way the result is more random.

antonderegt (changes requested):
Hey @jrazorx, you did a really good job, your code is very structured and you did error handling well. 🥳 🎉 

There are only two things that I want to ask you to change. Please have a look at the following.
At the bootom of [this page](https://thecsharpacademy.com/article/52) it says: 

> * Don't write too many statements without spacing. It makes your code harder to read.
> * On the other hand, never have more than one empty line. Or empty lines between code and brackets.

And on [this page](https://thecsharpacademy.com/article/58) it says:

> 3. Refrain from commenting
Code should be self-explanatory. Name your classes/methods/properties/fields so that you don't need to compensate by writing comments. Only write comments that are reminders for your future self or to let other developers know about aspects that the code can't possibly explain.

The way you name your variables and methods is already very good and descriptive. For example here, the comment is just repeating the method name:
```csharp
// Displays the menu
void DisplayMenu()
{ ... }
```

The reason for the above two things is explain in the introduction [here](https://thecsharpacademy.com/article/58). Also, I mean it as a compliment when I say that your code doesn't need all the comments because it's already well named. I always tell myself: comments should be about why I wrote a piece of code, not what the piece of code does.

Please update your code, then I'll be happy to approve it. You did well. 👍

MathGame.jrazorx/MathGame.jrazorx/Program.cs
jrazorx: Thank you for the tip ! Indeed, I looked into it, and it's better in terms of performance, without compromising readability :D
Fixed with https://github.com/TheCSharpAcademy/CodeReviews.Console.MathGame/pull/214/commits/d56d52e86562cd653ae1a9b384b421d09a374bd7

MathGame.jrazorx/MathGame.jrazorx/Game.cs
jrazorx: Fixed with https://github.com/TheCSharpAcademy/CodeReviews.Console.MathGame/pull/214/commits/c2325055faf2c379c71916807b8542f18e97b38d

antonderegt:
@jrazorx good job on the improvements! 🎉 
I'll approve it for now, but next time please focus on commenting less, your code is descriptive enough. It's important to follow coding conventions enforced by the company, in this case The C# Academy.

Good luck with the next projects! I hope I can review your code again, it was great to see you perspective, helps me learn as well.

jrazorx:
Thanks a lot for your review !

As the challenge suggested, I tried doing it on my own before looking at the video tutorials.

However, some of the videos covered more advanced stuff that did not get mentioned on freecodecamp certification lessons (Lists, Models, Enumerations, LINQ).

I'll update my code with this knowledge, and your suggestions

jrazorx:
Ah ! I did not expect you to approve it right away, I thought you would only review it again when I click on ask for a new review

antonderegt:
Haha, not everyone is aware of that option. Good that you are 😃

PR #219, opened 2024-04-10
Finishe Project

wfeliciano20:
I accidentally made 2 commits since I forgot to add my username to the folder


MathGame.wfeliciano20/Program.cs
```diff
@@ -0,0 +1,138 @@
...
+            userResponse = GetUserResponse();
+            Console.WriteLine($"The correct result is: ");
+            Console.WriteLine($"{firstNumber} * {secondNumber} = {result}");
+            score = validateResult(result, userResponse, score);
+
+            Console.WriteLine("\n");
+            break;
+        case 4:
+            Console.WriteLine($"{firstNumber} / {secondNumber} = ??");
+            result = mathGame.DivisionOperation(firstNumber, secondNumber);
```
VocalNight, line 69: Hey man, everything is looking good. The only thing need fixing is your division. if you try to do some divisions, you will see that the program is not letting you answer the questions since the 2 numbers being presented result in a division that is fractioned.

image


Could you double check your division logic in the code?

VocalNight (changes requested):

MathGame.wfeliciano20/Program.cs
wfeliciano20: this is happening since both numbers are generated from 1-100 but since the instructions said I could not show the results that had a fractional part, this is the predicament I found. Should I maybe limit the divisor to 1-10? so that more often there are non fractional results but It still isn'nt guaranteed that all divisor random numbers will result in integer division with a integer result. What do you suggest I could do to solve this if anything?

wfeliciano20: @VocalNight  I improved division now always shows a valid division

VocalNight (approved):
You did exactly what i would suggest. Everytime you enter division, make sure the numbers generated have 0 as remainer of the division, if not, generate again. Well done.

I will aprove the project, just one thing to keep in mind for future ones, as you can see, whenyou use the console a lot, the UI tends to get cluttered, try to think of a way to clean the console after certain operations to avoid that.

image


Again, good job and good luck on future projects!

MathGame.wfeliciano20/Program.cs
wfeliciano20: thank you

PR #220, opened 2024-04-10
Project finished

rankdjr:


MathGame.douglasfir_/MathGame/GameEngine.cs
```diff
@@ -0,0 +1,166 @@
...
+    }
+    internal MiniGame RunFixedLengthMode(QuestionType questionType)
+    {
+        DifficultyLevel difficultyLevel = util.PromptDifficultyLevel();
+        MiniGame game = new MiniGame()
+        {
+            Date = DateTime.Now,
+            gameType = GameType.Standard,
+            questionType = questionType,
+            difficultyLevel = difficultyLevel,
```
antonderegt, line 81: Hey @rankdjr at first glance the project looks good. But it doesn't run currently. It fails because `gameType`, `questionType` and `difficultyLevel` should be upper case. Did you do a refactor at the end?

antonderegt (changes requested):
Hey @rankdjr there's a small issue that prevents the project from running. Please check the comment I left in the code. 

I quickly took a look at the code and at first glance it looks very good 🥳 🎉 . Please fix the bug, then I'll do a proper review. But judging by what I can see now, it looks like you handled all the requirements and more!

MathGame.douglasfir_/MathGame/GameEngine.cs
rankdjr: Whoops. Looks like I may have added the wrong folder. Let me resubmit this with the correct project. Thanks for catching!

antonderegt:
Linked to #221

PR #222, opened 2024-04-14
Completion of MathGame Project

lilWe3zy:
Completion of Console math game project

I was very excited to get started and would appreciate any feedback you could provide.

This program has:
- difficulty selection
- question amount selection
- in memory history storage
- randomised operation mode


VocalNight (approved):
Hey. Good work on the project and congrats on doing all the challenges.

Honestly the only thing i can nitpick here is that your program class was a bit too big, if you could organize all those while loops in separated methods to keep it more organized it would probably make this perfect.

I also liked your use of tuples for the number range, not a lot of people use them.

I'll be approving the project, good job and good luck on future ones!

PR #225, opened 2024-04-26
Project Finished!

pinksparrows:


BBualdo:
Hello @pinksparrows, I will be your reviewer of this projects.

I can see that you uploaded built exe and dll files here. That's wrong approach, you should upload source code to review. I can't even open it, because for now I'm working on MacOS which doesn't support .exe files. 😅

Hit me up when you are done.

BBualdo:
Everything works as it should be.
Good job @pinksparrows !

I encourage you to try doing this challenge again without tutorial and maybe with challenges included 😁 This will really strengthen your knowledge!

Approved 😀

PR #226, opened 2024-04-29
Finished Math Game.

N-Endy:
I had a lot of fun doing this. I can't wait for your honest reviews.


MathGame.N-Endy/Services/QuestionService.cs
```diff
@@ -0,0 +1,81 @@
...
+                    question = GenerateQuestion("/");
+                    break;
+            }
+
+            return question;
+        }
+
+        public int CheckAnswer()
+        {
+            // Construct the arithmetic expression
```
antonderegt, line 62: Hey, comments like this are a bit redundant 😉

antonderegt (approved):
Hey @N-Endy very good job!! Works like a charm. The only thing that I would say is that some comments are a bit redundant and you should update to .NET 8 😆  .NET 7 will be losing support in two weeks! 😱

PR #227, opened 2024-05-04
Finished!

Ibraheem-Karim:
I used more functions and classes in comparison to the tutorial and changed the design of the whole program a bit. Please let me know if I'm overdesigning or overengineering it or if i did a good job, thanks!


Doc415 (approved):
Nice work. 
I am impressed by overriding tostring method in your GameRecord model.
Looking forward for your next projects.

PR #231, opened 2024-05-08
Adding my project for code review

Ccooper9893:
Most of the code is from the tutorial. The code that I would like reviewed would be the difficulty mode selection options. Maybe there is a better way to do it? Thank you!


Ccooper9893:
Thanks for the code review!

PR #236, opened 2024-05-16
commit Project (4/4 challenges)

ghost:
Thank you for all you do. I'd love to have your brutally honest feedback.  
Specifically I need feedback on how to partition / design my application so that it's more readable, understandable and expandable. I plan to start working in 14 months and I need all the advice on what I should focus on and cultivate.

### DIFF
--- a/Kamyar.MathGame/1.MathGame/Program.cs
+++ b/Kamyar.MathGame/1.MathGame/Program.cs
@@ -22,7 +22,7 @@ while (isRunning && numberOfQuestions > 0)
     }
     else
     {
-        Game game = new Game(questionTypeChoice, difficultyChoice);
+        new Game(questionTypeChoice, difficultyChoice);
         numberOfQuestions -= 1;
     }
 }


BBualdo:
Hello @kamyar-mazloom-1994. I will be your reviewer for this project. 

Generally your app is working fine, but there are some issues.

1. I like that you brought creative idea for number of questions. Creativity is important part of programming. But consider not closing application after user's "Rounds" are out. That's because user can't access Games History at any point in your application.

2. Consider Console.Clear() after some choices, because console quickly becomes messy 😅

Furthermore, I would advise you to dive deeper into classes and OOP basics 😁 For now you define Classes in Classes, which is not a good practice. You also write logic inside Class constructor, which is also bad practice. Classes are blueprints for objects, for example Game. Game should have properties like Score, Time, etc and methods like PlayGame(), StopGame() or something. And then, somewhere in your application you can create new instance of that class and you have access to everything you defined there:

```
Game game = new Game();

game.Score;
game.PlayGame();
```

It's crucial to understand that 😁 But for now just fix those 2 points I mentioned and you are good to go

PR #239, opened 2024-05-23
Math Game - 1st Submission

SCM42:
Thanks for all you do! Apologies if I somehow didn't do this right.

### DIFF
--- a/MathGame/GameEngine.cs
+++ b/MathGame/GameEngine.cs
@@ -1,0 +1,409 @@
+﻿using MathGame.Models;
+
+namespace MathGame
+{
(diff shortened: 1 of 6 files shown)


VocalNight (approved):
Everything is looking good!

Nice idea to separate the game history by difficulty level.

Just one nitpick i would give, is to take a look at your gameEngine class. It became very long since you have 3 difficulty levels for each operation. It's not necessary right now, but think of ways you could streamline it or make it so you have a global difficulty handler for the operations.

I will be approving the project. Good luck on your next ones!

PR #248, opened 2024-06-11
Finished Project

mtmattei:
This was a helpful first project. Be brutal and thank you.


TheCSharpAcademy:
@mtmattei Well done! Project approved! 😎✅

I suggest you try to solve the challenges as well. First without any help and then looking it up. These will take your learning to the next level 🤓

PR #250, opened 2024-06-12
Finished Math Game

jakubecm:
It was fun to play around with this. Didn't implement the difficulties because it would basically just mean generating numbers from different ranges.


TheCSharpAcademy:
@jakubecm Well done! It's great to see that completed some of the challenges. Those will take you to the next level 🤓 I always recommend you try these on your own before getting help 😎. I also like that you added summaries. Looking forward to reviewing your next project!

PR #251, opened 2024-06-14
Finished project

Lepros311:
Here's my first C# console project :)


TheCSharpAcademy:
@Lepros311 Well done! Project approved! 🥳 And well done for including the difficulty levels. Very good!

A couple of things you could have used: 
- Switch statements: they're a good option when branching execution based on several criteria. You could have used those in the menus.
- Methods: Methods make the code more organized and readable. 

Have a look into these for the next few projects!

PR #256, opened 2024-06-15
Finished project

SteveMuiyuro:
Hello Pablo! My project is done. For some reason the score is not printing at the end of every game but it 's saved on the history list. I've tried the debugger and indeed the code runs the console,writeline which is meant to display game over and correponding score but never shows ip on the console.


TheCSharpAcademy (changes requested):
@SteveMuiyuro Thanks for submitting! It looks like your project got picked up as a sub-module as the picture suggests: 
image

Try removing the .git folder from inside YOUR PROJECT's repository (not the academy's repository) and commit again. If it doesn't work, please create another PR, but this time before you move your folder into the academy's repository, remove the .git folder. Thanks!

PR #257, opened 2024-06-15
new files

Nightbringer1:


VocalNight (approved):
Everything is looking good.

The only thing i would point out to keep in mind in the future (and maybe fix it in this project when you can) is a way to clear the console after certain commands or operations. It can get quite a bit messy.

image

Other than that well done!
Good luck on future projects.

PR #261, opened 2024-06-17
Math Game Submission

Tanooki60:


TheCSharpAcademy (changes requested):
@Tanooki60 well done! Thanks for submitting! It's looking great but I found a bug:

image

Could you please fix it?

Also, I suggest you use only one method for all games. It's not a requirement but it will push your knowledge to abide to the DRY (Don't Repeat Yourself) principle. 🤓

Reach out if you have any questions. 👌🏻

TheCSharpAcademy:
@Tanooki60 Well done! ✅😎

I think there are a couple of tasks left before you can get your green belt. You can check the roadmap here: http://thecsharpacademy.com/dashboard/roadmap

PR #267, opened 2024-06-22
Project finished.

patrickbo89:


furiax (approved):
Nicely done, your first program works flawless.
Just only one small remark for future projects, keep an eye out for not using unnecessary white lines.
In Game.cs you have white lines at lines 8 and 11, in GameEngine.cs at line 6. These are not needed and we should avoid using them to comply to the code-conventions (https://thecsharpacademy.com/article/58/code-conventions).

But I'm approving your project :)
Keep up the good work !

PR #272, opened 2024-06-27
Finished mathgame

kAntonelli22:


TheCSharpAcademy:
@kAntonelli22 Well done! Project approved! ✅🤓 Well done on adding a timer, it's a shame you didn't complete the other challenges!

For the future we will get more strict about our code conventions, so make sure you read it and avoid. There were several instances of:

Unnecessary comments:
image

Camel case methods:
image

Also this is optional but in my opinion brackets are always a good thing for readability. I'd have a code block for this loop, for example: 

image

Looking forward to your next projects!

PR #275, opened 2024-06-29
Math Game with all challenges completed

JerryCJY:
Please be ruthless :)


TheCSharpAcademy:
@JerryCJY Well done! Project approved! Your app does all it's supposed to do.  And special congrats for having done all the challenges. I love that the difficulty is based on time as well 😁. Although you won't move as fast in the roadmap, you'll definitely learning more by doing that.

For the future, things to consider:
1 - There's a lot of code repetition. You want to abide to the DRY principle (Don't Repeat Yourself). You can aim for this in the next projects, but it would also be good exercise to try remove repetition from this project. I estimate you could half the amount of lines.

Examples: 

Your switch statement is bloated. You always want lean cases for better readability. You could create a bridge method that will handle all the console.writelines without having to repeat them:
image

You could have all games in one method. 
image

Overall great work and looking forward to the next ones!

PR #276, opened 2024-07-01
second pull

batus3010:


furiax (approved):
Well done, the different math operations work and you managed to implement two of the challenges. :)

However I do got two small remarks.
- You can make use of Console.Clear() to clear the screen after a menu choice, it looks much neater that way. After a couple of calculations the screen looks crowded now.
- You used a default case in the main menu, so when a faulty value is entered the user get prompted with a message the message to input a correct value. This is the way it should be, you did great here. But unfortunately you didn't implement this for the difficulty menu. Cause when I enter an other number it just continues as if I selected easy and if I enter something that isn't a number, the program crashes.

Since these remarks aren't listed in the requirements I'm approving your application, in future applications try to have some sort of validation on user input, you don't want your application to crash because of a faulty input. These validations will be required later on anyway so better to learn how to do them from the start ;)

Don't be discouraged by these remarks and keep up the good work.
And we hope to see some more applications from you in the future.

batus3010:
Hi @furiax. 
Thank you for the feedback and for approving my application! I'm glad to hear that most things were done right.

I'll definitely start using **Console.Clear()** to keep the screen clean after each menu choice. It makes everything look so much better. Also, thanks for pointing out the issue with the difficulty menu. I’ll make sure to add proper validation to avoid any crashes from incorrect inputs.

I really appreciate the constructive feedback. I’ll keep these points in mind for future projects. Looking forward to learning and improving more! 😃

PR #277, opened 2024-07-01
math game

echoblade100:


TheCSharpAcademy:
Well done! Project approved! Well done for completing all the challenges as well. I suggest you always do. It will help you retain your knowledge 😁

If I could suggest something for the future is trying to push yourself to abide to the DRY (Don't Repeat Yourself) principle. You could use one method for all games, since a lot of the code is the same. 🤓

Your code is very well organized. 👏🏻 I should point out that using 'this' isn't necessary in most cases.
image

I noticed when playing in difficulty 5 that many operands where in the 40-50 range. Was that a bug or the expected behaviour?
image

echoblade100:
Yes, this is expected behavior, at level 5, pairs from 41-50

PR #278, opened 2024-07-01
Kriz-J MathGame

Kriz-J:


TheCSharpAcademy:
@Kriz-J Well done! Project approved! Well done for completing all the challenges as well. I suggest you always do. It will help you retain your knowledge 😁

If I could suggest something for the future is trying to push yourself to abide to the **DRY (Don't Repeat Yourself)** principle. You could use one method for all games, since a lot of the code is the same. 🤓

Your code is very well organized. 👏🏻 I should point out that using 'this' isn't necessary in most cases. 
image

I noticed when playing in difficulty 5 that many operands where in the 40-50 range. Was that a bug or the expected behaviour? 
image

TheCSharpAcademy:
@Kriz-J Well done! I gotta say, one of the best Math Game projects I've seen. ✅🤓 I love all the UI elements you added to it. And nice use of inheritance for the games! Looking forward to reviewing the next ones.

PR #279, opened 2024-07-02
Project Finished!

rholexxdizon:
Had so much fun doing the Math Game! Learned a lot! Looking forward for the brutal code review.


TheCSharpAcademy:
@rholexxdizon Well done! Project completed! ✅🤓 And well done for completing challenges! I recommend you always do that. It's the best way to retain knowledge and get better at problem solving. 👌🏻

For the future, try to think of ways to avoid code repetition. It's what we call the DRY principle (Don't Repeat Yourself). You could have just one method for all games for example, and a method to determine the numbers based on the difficulty, so you didn't have to repeat it here:
image

Overall very good! Looking forward to seeing your next projects. 😊

PR #280, opened 2024-07-02
Project finished!

kp861:
Hi, although I have no doubt could have done better here, however it was still fun. I am last year CS student and I have done myself wrong for not practicing more apart of what was required :) Although this code is terrible, I hope that during practicing here, I will improve myself quickly!

### DIFF
--- a/MathGame.sln
+++ b/MathGame.sln
@@ -1,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.10.34916.146
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MathGame", "MathGame\MathGame.csproj", "{630E594E-A743-4783-A275-870FE04C6DAC}"
+EndProject
+Global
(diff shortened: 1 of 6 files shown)


TheCSharpAcademy:
@kp861 Well done! Project approved! You completed all requirements ✅😊 

The code isn't bad, but there are things that can be improved  for your next project. Here are few things.

1 - You should remove unused 'usings'. You left them in all files. You can also add a semi-colon after the namespace to remove one level of indentation (i.e. namespace MathGame;) And line 7 is an unnecessary empty line. 
image

2 - You can look into string interpolation to avoid ugly plus signs: 
image

3 - Your code would be more readable if you outsourced it into methods. Otherwise your branching becomes very messy and hard to read. 
image

The code above could be : 
``` 
if (condition) 
{
     DoSomethingMethod();
}
else 
{ 
     DoSomethingElseMethod();
}
```

4 - There's a lot of code repetition in GameMechanics. You could have all games in one method, using variables to change the logic conditionally. This would follow the DRY (Don't Repeat Yourself) principle. 

5 - Always name things so that it's easy for a 4-year old dev to understand what's going on. **"bool r**" is a head-scratcher. On the other hand, "**bool myVeryVerboseButClearlyUnderstandableVariable**" makes everyone's lives easier. 😁

PR #281, opened 2024-07-06
Mathgame

mrmyxc:


csa-maths-game/csa-maths-game/Game.cs
```diff
@@ -0,0 +1,145 @@
...
+            return (correct, span);
+        }
+
+        private int GetUserResponse()
+        {
+            int response;
+            bool success = false;
+            do
+            {
+                success = Int32.TryParse(Console.ReadLine(), out response);
```
javedkhan2k2, line 91: I can set response to 0 and negative values. When I set the number of rounds to -5, I could not able to play the game anymore.
Kindly add the validation so that the number of rounds should be greater than 0.

csa-maths-game/csa-maths-game/Game.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        public void Play(IOperation op)
+        {
+            Operation = op;
+            bool correctAnswer = true;
+            TimeSpan lastTimeTaken;
+            TimeSpan totalTimeTaken = TimeSpan.FromSeconds(0);
+            int score = -1;
+
+            while (correctAnswer && score < NumberOfRounds)
```
javedkhan2k2, line 22: Can you change the logic so the game is not quit when the User give a wrong answer.
This is also A valid game but this was a requirement that user need to answer predefined number (in your case it is max int) questions and in challenge one of the part was to make it dynamic (which is already implemented).

csa-maths-game/csa-maths-game/Division.cs
```diff
@@ -0,0 +1,24 @@
...
+        {
+            return "Division";
+        }
+        public char GetSymbolChar()
+        {
+            return '/';
+        }
+        public bool IsValidOperands(int operandA, int operandB)
+        {
+            if (operandA == 0 || operandB == 0) return false;
```
javedkhan2k2, line 19: when operandA has 0 value, it is a valid mathematical operation. You can change it or leave it.

javedkhan2k2 (changes requested):
@mrmyxc a job well done.
Please review the comments and fix the issue with Number of rounds when the user set it to negative values.
Don't forget to commit.

csa-maths-game/csa-maths-game/Game.cs
mrmyxc: thanks, done

csa-maths-game/csa-maths-game/Game.cs
mrmyxc: I didn't see this requirement to answer a predetermined number of questions. 
Was it in the tutorial video? because I didn't watch it

javedkhan2k2:
@mrmyxc Approved. Brilliant work.
I will only suggest to convert part of your code to classes as well e.g.
`static List<(int score, string name)> highScores = new List<(int score, string name)>(MaxScores); `
define a class
`static List<GameScore> highScores = new List<GameScore>(MaxScores);`
Good luck with your next Project.

PR #286, opened 2024-07-11
Project finished!

NgocAnhDo26:


TheCSharpAcademy:
@NgocAnhDo26 Well done! Project approved! 🥳🥳 And special congratulations for having done the challenges! 👌🏻👌🏻

A few things I could point out:
1. Don't forget to tidy up and remove unnecessary comments/usings: 
image

2. For readability, it's good to keep your switch cases leaner. Outsource the code to a method. Also, there's a bit of repetition. You could dry it down to one method that would cater for all cases, following the DRY (Don't Repeat Yourself) principle.

3. You could use a default case in the switch statement to handle validation, which would make this unnecessary: 
image

4. I like how you used Inheritance with the games, but in each game, again there's a lot of repetition. You could have a method handling all games.

But overall well done! Looking forward to the next ones! 

PS: You need to submit your Free Code Camp Certification to get your Green Belt. It would also be good if you joined our Discord community 😎

PR #294, opened 2024-07-21
naina35 MathGame project

naina35:
Completed MathGame project

### DIFF
--- a/MathGame.naina35/MathGame.csproj
+++ b/MathGame.naina35/MathGame.csproj
@@ -1,0 +1,10 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+</Project>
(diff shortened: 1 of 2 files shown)


TheCSharpAcademy (changes requested):
@naina35 Hi there! Thanks for your first submission! 🤓 The program works well, but I found a couple of strange behaviors:

1 - After you prompt "Guess the answer for....", I input a number and I get asked to enter my choice, which means I always have to input my answer twice.

2 - The History always says I answered 0 questions.

image

Please make the following corrections:
1 - User should only input once.
2 - Choosing previous history should show a **List of Games**.
3 - For better readability, clear the console after a game is played

Thank you! You're close 👏🏻👏🏻

naina35:
In my program:
In one game you can play several questions of the four operations and can choose to display the menu/history anytime
Although after you end the game or start a new one the previous history is deleted.
I am confused on how i should change the program... From what i understand:
**I can change that to record the history of ALL games played in current program**
**Once user enters a choice questions should only be presented of that operation in current game**
For example when it was presented you answered 2 questions.... you could have chosen also to display history 
. 
image

TheCSharpAcademy:
Makes sense. Thanks for the explanation. I'll approve now 😎. For future projects:

1 - Clear the console after an operation for better user experience
2 - Present the menu again, so the user doesn't have to guess what's going on.
3 - It would have been better to have the history persist across multiple games.

Looking forward to the next ones!

PR #299, opened 2024-07-24
Waiting for review

KroksasC:


TheCSharpAcademy (changes requested):
@KroksasC Hi there! Thanks for submitting! Unfortunately your project is in .NET Framework, which is the legacy version of .NET. Could you please create the project using .NET instead? It would be a lot of effort to install .NET Framework just for reviewing one project 😟

Also make sure you remove unused usings. 😎
image


Thank you!

PR #304, opened 2024-07-31
Finished project

przemyslaw-lewandowski:


TheCSharpAcademy (changes requested):
@przemyslaw-lewandowski Hey there! Thanks for submitting and well done for completing the challenges! 👌🏻

🔍I found a couple of issues. When running a division, I don't see a sign. And also not fulfilling the requirement below. Could you please fix it? 
image

image.

🔴In the menu you should add a default case to your switch statement so the app doesn't break if I don't type 1-7: 
image


🔴Also, optionally, you could try to improve this to abide to CLEAN principles: In the Operations and Difficulty classes there's a lot of code repetition. You could use one method and just change what's necessary based on parameters.

przemyslaw-lewandowski:
Fixed the issues (hopeffuly). When in comes to division, dividends from 0 to 100 are set only for easy difficulty.

TheCSharpAcademy:
@przemyslaw-lewandowski Well done! Project approved! ✅👌🏻And well done for completing the challenges. I recommend you always do 😁

Another example of where you could go a bit more DRY. Notice how you have a bit of repetition here? You could have created a method that would take the difficulty ranges so you didn't need to repeat all of this code. I suggest you always keep an eye of these opportunities to keep the functionality with less code 🤓.

image

Join Us On Discord
Oh and if possible join our Discord Community and add your username [to your profile](https://thecsharpacademy.com/dashboard/profile) so you can be part of the community as well. 😁

PR #308, opened 2024-08-04
My First project

Eyad-Mostafa:
I have learned a lot, thank you <3


TheCSharpAcademy:
@Eyad-Mostafa  Well done! Project approved! ✅🤓And well done for completing the challenges! I recommend you always do 🤓

One thing I could say is that you could have used just one method for all games. That would abide to the DRY principle (Don't Repeat Yourself). If I understand well, you followed my tutorial and that's how I created the solution but it's something to keep in mind for the future 😁

Looking forward to seeing your next projects! 🤓

PR #350, opened 2024-08-29
Project finished

BonginkosiShongwe:


chrisjamiecarter (approved):
Hey @BonginkosiShongwe ,

First of all, great job on your Math Game application! 🎉 It’s clear you’ve put a lot of effort into this, and you’re making great progress.
- Well done on structuring your solution in the correct way!
- Great use of Interpolated strings.

I realize that much of your code likely follows the tutorial, but I have a few suggestions that I believe can help you enhance it even further:

**Style Guide**
Not required, but when you have written code or finished a tutorial have a look at [Microsoft's style guide](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions) to align with best practices.
Here are some pointers:
- Use file-scoped namespaces to save an indentation level
```
// Block-Scoped Namespace
namespace MathGame.BonginkosiShongwe
{
    // Code here.
}

// vs

// File-Scoped Namespace
namespace MathGame.BonginkosiShongwe;
// Code here.
```
**Variable Naming**
- Try not to name your variables according to their data type. For example, `string sQuestion`, `int iAnswer`, `int[] inumerator` and `int[] idenomenator`. What happens if you update the type from int to long? This will make your code more maintainable.

**User Experience**
- Why don't you make the menu options case insensitive?
- Why did you make an invalid option close the app? Why not display the invalid option message, but then go back to the main menu?

**Bugs**
Unfortunately, I did spot a bug within your application:

1. View Previous Games Bug:
It seems that when you have done multiple previous games, and then press `V` to view them, only the first game you played is shown. 
NOTE: you call `prevGame.Add(sQuestion);` in the `Addition()` method only. `prevGame.Append(sQuestion)` in the `Subtraction()` method. And nothing in `Multiplication` or `Division`.
NOTE NOTE: `list.Append(x)` does not do the same as `list.Add(x)`. Add(x) adds the item to the current list. Append(x) adds the item to a new list which is returned. So you would have to do something like `list = list.Append(x)`. Hopefully this makes sense.

I suggest revisiting to try understand and fix any issues, but I don't think they will block me approving this PR..

Feel free to reach out if you have any questions or if you'd like to collaborate further on any of these points.

I will go ahead and mark as approved, keep up the excellent work on the next projects! 😊

Best regards,
@chrisjamiecarter

PR #371, opened 2024-09-26
Finished MathGame

Behram-Khan-1:
needs polishing

### DIFF
--- a/MathGame/EnumOptions.cs
+++ b/MathGame/EnumOptions.cs
@@ -1,0 +1,22 @@
+
+
+public enum EOptions
+{
+    Addition = 0,
+    Subtraction =1 ,
+    Multiplication=2,
+    Division=3,
+    Difficulty=4,
+    History =5,
+    End=6
+    
+}
+public enum EDiffuculty
+{
+    Easy =0,
+
+    Medium,
+
+    Hard,
+}
+
--- a/MathGame/MaintainHistory.cs
+++ b/MathGame/MaintainHistory.cs
@@ -1,0 +1,27 @@
+
+public static class RecordKeep
+{
+    public static List<string> History = new List<string>();
+
+    public static void RecordData(string fn, string sn, string ans, string symbol )
+    {
+        History.Add(fn);
+        History.Add(sn);
+        History.Add(ans);
+        History.Add(symbol);
+    }
+    public static void GetHistory()
+    {
+        if(History.Count == 0)
+        {
+            Console.WriteLine("No history to show");
+            return;
+        }
+        for (int i = 0; i < History.Count ; i = i+4)
+        {
+            Console.WriteLine(History[i] + "  " + History[i+3] +"  " +  History[i+1] + " = " + History[i+2]);
+        }
+    }
+
+
+}
--- a/MathGame/Program.cs
+++ b/MathGame/Program.cs
@@ -1,0 +1,121 @@
+﻿
+public static class MathGame
+{
+    static void Main()
+    {
+        bool result = false;
+        bool optionsSelected = false;
+
+        while(!optionsSelected)
+        {
+            EOptions mainmenu =  MainMenu();
+            do
+            {
+                if((int) mainmenu < 0 || (int) mainmenu > 6)
+                {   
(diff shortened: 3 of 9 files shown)


MathGame/MaintainHistory.cs
```diff
@@ -0,0 +1,27 @@
+
+public static class RecordKeep
+{
+    public static List<string> History = new List<string>();
+
+    public static void RecordData(string fn, string sn, string ans, string symbol )
```
baruchvlz, line 6: What is `sn`? Ideally you would want to use variable, or argument names in this case, that can make it easy to understand what it is.

MathGame/MaintainHistory.cs
```diff
@@ -0,0 +1,27 @@
+
+public static class RecordKeep
+{
+    public static List<string> History = new List<string>();
```
baruchvlz, line 4: I wouldn't make this public since you're using a method as a getter.

MathGame/Program.cs
```diff
@@ -0,0 +1,121 @@
...
+        Console.WriteLine("1 - Subtraction");
+        Console.WriteLine("2 - Multiplication");
+        Console.WriteLine("3 - Division");
+        Console.WriteLine("4 - Set difficulty");
+        Console.WriteLine("5 - History");
+        Console.WriteLine("6 - End");
+        
+        string selectedOption = Console.ReadLine();
+
+        int select = int.Parse(selectedOption);
```
baruchvlz, line 82: Using `int.Parse` will crash your program with an unhandled exception if it cannot parse. I suggest you use `int.TryParse`, this way you can also check if the parse was successful:

```c#

bool selected = int.TryParse(selectedOption, out int parsedSelection);

if (!selected)
// display error
```

Alternatively you can use it in a do/while loop since `int.TryParse` returns a boolean.

```c#
string selectedOption;
bool selected;
int parsedSelection;

do {
    selectedOption = Console.ReadLine();
    selected =  int.TryParse(selectedOption, out int parsedSelection);
} while (!selected);
```

(Please not that this is pseudo code, it might not work but it's only here to demonstrate what it can do)

MathGame/EnumOptions.cs
```diff
@@ -0,0 +1,22 @@
...
+}
+public enum EDiffuculty
+{
+    Easy =0,
+
+    Medium,
+
+    Hard,
+}
+
```
baruchvlz, line 22: What does the `E` prefix mean?

MathGame/Program.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        Console.WriteLine("Pick a number to continue...");
+        Console.WriteLine("0 - Easy");
+        Console.WriteLine("1 - Medium");
+        Console.WriteLine("2 - Hard");
+    
+        string selectedDifficulty = Console.ReadLine();
+        int diff = int.Parse(selectedDifficulty);
+        eDiff = (EDiffuculty)diff; 
+            if(diff >= 0  && diff < 3 ) exitCondition = true;
```
baruchvlz, line 107: This is missing 1 tab indentation since it's inside the `while` loop. This makes it easier to understand the scope of the loop.

README.md
```diff
@@ -0,0 +1,7 @@
+# Math-Game-C-
+ 
+My Worst Coding attempt ever lmao.
```
baruchvlz, line 3: I don't agree with this statement, it shows understanding of basic fundamentals of any programming language (loop, variables, conditionals, etc).

MathGame/EnumOptions.cs
Behram-Khan-1: enum

TheCSharpAcademy:
@BlessedBarrel Have you submitted this project in the [Academy's dashboard](https://thecsharpacademy.com/dashboard/submit-project)? I can't see it there. 😟

TheCSharpAcademy:
@BlessedBarrel Closing the PR for Admin purposes, feel free to reopen when you submit in the [Academy's dashboard](https://thecsharpacademy.com/dashboard/submit-project). 😄

PR #381, opened 2024-10-12
Project finished!

GreyKlouds:
I implemented the difficulty level challenge.


arashi256 (changes requested):
Hi there! It appears you have only uploaded binary compiled files. While this works, I would like to see the project source code! It is code review, not app review :D  
Some things:-
- The time, on the greeting only displays as "00:00:00" not the actual time.
- Please upload your project named like "MathGame.[Your user name]" format rather than "FirstProject"
I have tested all the functionality and everything works as intended, including input validation and difficulty levels, aside from that time display, great job. The calculation history also works perfectly.
Upload the project code files and fix that error with the time and I will approve it! :)

PR #393, opened 2024-10-31
MathGame - entire project

awons:


TheCSharpAcademy:
@awons Well done! Project approved! 👌🏻🤓✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster.

🔍**One Piece of feedback:** Your code is super well organized and you seem to care about architecture, which I really like. However it seems a little over engineered. You created 12 classes for the operations! You could have a single class with a single method that could handle all the games. Although this is a matter of personal preference, I suggest you lean towards simplicity in the future. 😄

**A couple of examples:** 
1️⃣ It defeats the purpose of a method to simply return the result of a statement. 
image

2️⃣Why so many wrappers for a simple statement? Unless this is used everywhere, just keep the statement itself where it's needed:
image

🗺️Don't forget you need to submit your Free Code Camp certificate to get your 🟩🟩 Olive Green Belt. [You can see the requirements for each belt here](https://thecsharpacademy.com/dashboard/roadmap). 

Oh and if possible join our Discord Community and add your username [to your profile](https://thecsharpacademy.com/dashboard/profile) so you can be part of the community as well. 😁

Looking forward to seeing your next projects! 😄

PR #395, opened 2024-11-01
yay i finished it!

lynnthegrinch:
my attempt at the math game!


TheCSharpAcademy (changes requested):
@lynnthegrinch Thanks for submitting! There are a couple of requirements not being met. With the divisions I keep getting this. The program should be able to loop through divisions until it spits out a good one. 

image

Also games history should be a list:
image

Don't hesitate to reach out if you need help 😄

lynnthegrinch:
working on it ! i just fixed the div problem and im going to work on the array system again, thank you!

PR #396, opened 2024-11-03
My first completed Project!

fdv99:


arashi256 (approved):
Hello! I liked the fact that your score for each run is calculated as a percentage especially and everything meets the requirements. However, Your input validation needs some extra attention - as per the screenshot (hopefully) attached, I was able to enter a level beyond 10-100 which seems to crash the session. I was also able to enter my name as either nothing or just numbers.
I am going to approve this because technically speaking you've met all the requirements (nice, nice) but input validation is like 50% of making apps stable and solid and it's going to be very important soon for you in the roadmap!

shot

But this observation is more to help you prepare for the next projects not a complaint of this project. You did good 👍 See you in the next project

PR #421, opened 2024-11-24
Implement challenge requirements except for voice input

Alvind0:
Voice input is too difficult at the moment so I will skip it and come back later when I'm better skilled.


TheCSharpAcademy:
@AshtonLeeSeloka Closing this as we unfortunately don't review again after the project is approved, but well done for going through the feedback and improving your app! You should definitely have it in your own Github portfolio 🫡🫡

PR #449, opened 2025-01-03
Project finished

DenisN00:


hasona23 (approved):
Hi Mr. @DenisN00 
Hope you are fine. Congrats 🎉  on Finishing your first challenge "Math Game". Keep up the good work 🔥

I just have few comments for future projects and I hope they help you
1. Your code is really well structures with functions and no reptition (Except in point no. 4 ill talk about this). Anyway it would have been even better if you can separate them into files atleast another file called MathOperations where you'll add the 4 Math functions.
2. Line 8: Change String array (gameHistory) to List with capacity 100 instead
```List<string> gameHistory = new List<string>(100). Why?
    1- To prevent if i played too much the app crashes. Arrays cant expand beyond the 100 elements you set but in list it would just increase capacity. {Note: The current part is just advanced if you don't understand it don't worry it is just a matter of time. } 
ok so when i pass bounds of array like 101 elements the program would crash but when i make a list even as ```gameHistory = new List()``` without 100 if you passed bound it has an underlying array which would just be turned to an array of 200 elements ```string[200]``` instead of crashing.

3. Line 48: What is Case 4. why isnt it present in the Menu options. if 
for testing purposes then make it clearer with comments. I don't understand what is it basically for

4. In CheckResult function. 
It can be Written as: 	
```if (correctResult == userResult)
    {
        Console.ForegroundColor = ConsoleColor.Green;
        Console.WriteLine($"{userResult} is the correct result!");
        gameHistory[gamesCounter] = $"What is the result of {x} {operationType} {y} ? \nYour result was correct.";
    }
    else
    {
        Console.ForegroundColor = ConsoleColor.Red;
        Console.WriteLine($"{userResult} is a incorrect result.");
        Console.WriteLine($"The correct result is: {correctResult}");
        gameHistory[gamesCounter] = $"What is the result of {x} {operationType} {y} ? \nYour result was false.";
    }
    gamesCounter++;
    Console.WriteLine("\nPress any key to continue . . .");
    Console.ReadKey();
    Console.ForegroundColor = ConsoleColor.Gray;
    Console.Clear();
    return correctResult == userResult;```
This is just to prevent repetition but it is not important perhaps just a prefernce

anyway I'm glad that the project was done this way especially the change in colors part which you'll do even more in further projects using a library called Spectre.Console

Thank you 🥇 .

DenisN00:
Hey,

Thanks a lot, and great review!

I haven't learned about lists in the C# Foundationals Course but that seems to be a good usecase since the games history is dynamic. Im definetly gonna check out how that works and try to include it in my next project when necessary.

Case Nr. 4 I forgot to remove :(

The restructured CheckResult function is a good point. I really try to go after the best practice principle as much as possible. - I was sloppy here.

I really appreciate the review. That's a big help and motivation for me !!!

PR #478, opened 2025-02-17
S1m0n32002 - MathGame - V 1.0.0

S1m0n32002:


TheCSharpAcademy:
@S1m0n32002 Project approved! 😄✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster. I can see you have previous coding experience? You have very good standards 🤵🏻

🔍️The only piece of feedback I think I can give you at this stage is that in .NET, the Program.cs class and the Main method are used by convention as the starting point of your app. So you can set things up in there but the actual logic usually goes into its own classes. 

S1m0n32002:
Thanks! Yes, as I said on the discord I already have a job where I must use vbnet.
Btw I'm S1m0n3#5689 on discord 🇮🇹

PR #484, opened 2025-02-25
Complete math game

jamiemetca:
All additional challenges (minus Azure AI) complete and ready for review.


TheCSharpAcademy:
@jamiemetca  Project approved! 😄✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster. 💪🏻

🔍The project looks good, it's well organized. The only feedback I have is that there seems to be some signs of over engineering. Example. Why create methods for these one liners? All of the sudden we one line becomes four. As far as I'm concerned these wouldn't need a method.
image

On the same token, why do we have a different project for the models? 🤔 I'd recommend only creating different projects when it makes sense to do so. Of course, it's not wrong, it comes down to your preference. But for such a small project, a different folder would be enough. Also, a lot of the classes inside the "models" projects are actually business logic, so it shouldn't be under the "models" umbrella.  I hope it makes sense?

👪Oh and if possible join our Discord Community and add your username [[to your profile](https://thecsharpacademy.com/dashboard/profile)](https://thecsharpacademy.com/dashboard/profile) so you can be part of the community as well. 😁

Looking forward to seeing your next projects! 🤓

PR #535, opened 2025-06-09
finished project

KacperZys:
Cool project. I had much fun doing it!


TheCSharpAcademy:
@KacperZys  Project approved! 😄✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster.

🔍️In terms of feedback, a few things:
1 - Ideally your game should have a few questions instead of just one, for better user experience. 

2 - As per the example below, every time you use Convert.ToInt32, your app will crash if the user doesn't input an integer. Ideally you'd handle those cases. That's what we call input validation.
image

I'm still approving since validation isn't a requirement, but keep these things in mind for the next projects!

PR #546, opened 2025-07-08
request

mosnyan:


TheCSharpAcademy (changes requested):
@mosnyan Thanks for submitting and apologies for the delay in reviewing 😟. It looks great! 

I'm not overly versed in design patterns but it seems like you're using the "Strategy Pattern", was this a choice for learning or you have used it professionally? You said it was over engineered but I disagree. You need to treat these projects as sandboxes where you can try new things, that's how we just give you high level requirements 😎

Before I can approve, there's just one requirement missing. At this stage you don't let the users view a history of played games. Also, ideally the game would run in a loop so it doesn't stop after each game.

image

mosnyan:
d’oh, that will teach me not to pay attention

i’m not a professional software dev, i actually work in industrial automation as a controls specialist but there is a thing we call « it/ot convergence » where the two fields gradually develop a level of overlap

i ended up enrolling back in university to study cs in order to boost my career a bit and be in a good position to manage this overlap. but my local uni leans on the theoretical side so aside from small programming assignments in most classes, there are really only a handful of courses where the scope of the assignments is somewhat substantial in between studying automata theory and doing typing tree proofs.

i don’t really have any ideas for projects outside of job stuff, and i don’t have a mentor either, so i really appreciate what you’re doing here with your c# program. i love that it’s not code along - that’s not how you learn IMO - but there is a feedback system all the same in these code reviews.

programming industrial machinery is not the same as « normal » programming. as silly as it sounds, OO paradigm is basically the hot new thing for us so while imperative logic is no trouble for me, all the more architectural/design/pattern stuff is a bit more foreign on the OO side. so i try to have fun with the assignments when it makes sense

anyway i will have this ready in the next few days

mosnyan:
hello,

as requested, i have added the history function

TheCSharpAcademy:
@mosnyan  Project approved! 😄✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster.

🔍️In terms of feedback, just a couple of things: I noticed some signs of over engineering. Example: Why create a display method that does only one thing, which is Console.WriteLine()? Since this is a one-liner, is it necessary to wrap it with another method? 

image

image

Also you don't want to use exceptions for validation. Exceptions are for .... exceptions. Validation errors are expected, so they don't qualify as such. A simple switch case with a default case would handle this.

image

Other than that, everything looks good. The code is well organized, it follows DRY and SOLID and I like how you set "strategies" for different games when needed.

👪Oh and if possible join our Discord Community and add your username [to your profile](https://thecsharpacademy.com/dashboard/profile) so you can be part of the community as well. 😁 😊 And do you mind choosing a country in the academy? I think it looks prettier with the flag. If you're not comfortable with that you can choose a random country 😎

mosnyan:
thank you very much for the review

so my thinking for the one-liner method is definitely over-engineering - what if i want the view part not to be a console but instead some kind of other desktop or web app. i need a method to generically communicate with the view, like show text in a textbox or popup instead of printing to console.

i saw that one of the projects was the math game but in .net maui so i thought i'd save myself some pain later and attempt to have to just rewrite the view (ideally i would have made a view interface with these methods), but turns out that since i run linux at home it'd be hard to do the maui projects.

incidentally, if i submit the .net maui projects using avalonia instead of maui, would it be possible to still get code reviews?

PR #552, opened 2025-07-24
Requesting project review

mxrt0:


MathGame/Program.cs
```diff
@@ -0,0 +1,100 @@
...
+                    while ((isFirstNumberValid = int.TryParse(Console.ReadLine(), out firstNumber)) == false)
+                    {
+                        Console.WriteLine("Please enter a valid integer!");
+                        Console.Write("Enter 1st number (integer): ");
+                    }
+
+                    Console.Write("Enter 2nd number (integer): ");
+                    bool isSecondNumberValid;
+                    int secondNumber;
+                    while ((isSecondNumberValid = int.TryParse(Console.ReadLine(), out secondNumber)) == false)
```
VocalNight, line 51: This is repeated twice, could have been made into a method

VocalNight:
This is pretty good!
I liked what you did, using a dictionary to map out the operations and just retrieving them.

I would have liked the check if the number input is valid or not to be a separated method, but that's mostly a nitpick.
The project would be even nicer with a Readme, but that can be added later.

Overall this is very solid, good job and keep up!

PR #560, opened 2025-08-11
Added all math game types

NemisIGuess:
· User can select difficulty
· User can see past games history


TheCSharpAcademy:
@NemisIGuess Project approved! 😄✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster.

🔍️In terms of feedback, just a couple of things: You can clean up the screen more often and add more spacing for better readability. It gets cluttered pretty quickly.

image



🔍 A dictionary is probably an overkill for this. You could just use an enum. Also, if you're going to declare a dictionary, you could do it in 6 lines instead of 20 plus. No need to escape every bracket 😊

image



🔍This is a clear example of unnecessary methods. There's no need to create a method for a one liner. If all your method does is to call Console.WriteLine(), why not just call Console.WriteLine(); directly? 😎

image

PR #574, opened 2025-09-14
My contribution

rafa-alves6:
I see some room for improvement, I could add some comments to document it but it looks quite readable. 
Looks Good To Me™

### DIFF
--- a/MathGame.csproj
+++ b/MathGame.csproj
@@ -1,0 +1,10 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net9.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+</Project>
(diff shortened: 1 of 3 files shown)


TheCSharpAcademy:
@rafa-alves6  Project approved! 😄✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster.

🔍️In terms of feedback, just a couple of things: You could have made a game with more questions where the users don't have to pick an operation after each question. Also, in your switch case, you didn't need to parse into an int. It could still be a string, and any string other than the options given would hit the default. So the lines below weren't necessary at all:

image

Also, 8 spaces of indentation is too much. The convention in C# is 4. [You can read our code conventions here](https://www.thecsharpacademy.com/article/58/code-conventions)🤓. There are vast unnecessary empty spaces in your code due to the multiple levels of indentation: 

image

Other than these small issues, the project was great! 

rafa-alves6:
Thanks for the feedback! 

I'll look into ways to ask for more questions, I think I'll do 3 at a time.

As for the first Parse, the rationale behind it is:
- if the user inputs a string, say, "divide", he would be instructed that only numbers are supported, then:
- if he inputs a number that isn't in the supported range, that would fall onto the default check.
but what you said makes sense too!

And about the indentation, I remember I saw Linus Torvalds comment about using 8 spaces/tab, and I tried it out here. 
I'll adhere to the conventions from now on!

Also, I'd like to ask if you would say this code complies with DRY principles. I tried to do that by using the AskQuestion() function, but I thought it would be better to break the code up in smaller parts.
I just don't know if it is very DRY-ey, cause there were some repetitive parts.

Once again, thanks for the feedback.

PR #575, opened 2025-09-14
Final Math game

rafa-alves6:
Contains challenges 'Random game' and 'Timer'. I'd like to ask if complies with DRY principles.

Changed to adhere to the [C# Academy Code Conventions](https://www.thecsharpacademy.com/article/58/code-conventions).

### DIFF
--- a/MathGame.csproj
+++ b/MathGame.csproj
@@ -1,0 +1,10 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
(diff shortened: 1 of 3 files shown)


TheCSharpAcademy:
@rafa-alves6 Looking good 👏. Usually we don't review after it's been approved so don't worry about submitting a new PR. Our suggestions are for your own self-study so you can apply what you've learned in future projects.

One thing. The method below seems unnecessary. There's no need to create a method if it contains a one-liner. You can simple use that one line instead 😎.

image

PR #577, opened 2025-09-29
Math Game

Esaiy:
cool little project


TheCSharpAcademy:
@Esaiy  Project approved! 😄✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster.
 
🔍️In terms of feedback, just a couple of things: I highly recommend you learn about the [DRY Principle (Don't Repeat Yourself) ](https://www.thecsharpacademy.com/article/30006/dry-principle-csharp). It's the most important challenge in this project. Your code is neatly organized but there's a little bit of repetition for the games. You should be able to have all games using just one method.

Also you could have used switch statements instead of so many if statements to make your code more readable. While your approach isn't wrong, it could have been a little more organized. If you use a switch statement, you wouldn't need to parse the user input into an integer. Anything other than the correct inputs could be handle by the default case. I suggest you explore this for the next console projects 🤓

But overall great work! 👏👏

PR #580, opened 2025-10-04
Project finished!

H4rkonX:


TheCSharpAcademy:
@H4rkonX Project approved! 😄✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster.

🔍️Overall, great work! Just a couple of pieces of feedback: 

1️⃣ Have a look at our style guide since there are a few violations in your code. e.g. Lots of unnecessary comments and unused usings (you can see in the picture below)[You can read our code conventions here](https://www.thecsharpacademy.com/article/58/code-conventions)🤓.

2️⃣You're using exceptions to handle validation. That's not good practice since exceptions are "expensive" in terms of the use of resources. You want to leave exceptions only for ...... Exceptions! 😆. None of the code below is necessary, you could simply handle the wrong input in a switch statement default case, which you left empty.

image

image

PR #584, opened 2025-10-09
Update to add solutions for other challenges

Eulysses001:
Added Stopwatch for timer.
Added Random Game Mode and Difficulty.
Used DRY Principle (not sure if fully utilized).


TheCSharpAcademy:
@Eulysses001 Well done for reworking your project. Unfortunately we don't review projects again after it's been approved (too many projects!), but it's great that you tried to improve it. I hope the advice was useful and you apply what you've learned in the next projects 🤓

PR #586, opened 2025-10-13
MathGame Console Finish

Myhos0:


TheCSharpAcademy:
@Myhos0 Project approved! 😄✅ And well done for completing some of the challenges! I recommend you always do that, as it pushes you and makes you learn faster. I also really like that you added color to your app and that the games have a top bar so you can keep track of the scores. Well done! 👏

🥋And congratulations! You got your Green belt!! 🟩🟩

🥋You're only two projects away from getting your Olive green belt!! 🫒🫒 The habit tracker and the coding tracker 👏

🔍️In terms of feedback, just one thing got my attention. None of the methods in the picture below are necessary. You don't need to wrap a one-liner in a method, it's already concise enough and it's a clear example of over-engineering. Instead just call the method directly. 

image

🕵️‍♀️Also, in the same screen shot there are a code tidiness issues: e.g. unused usings, and unnecessary empty spaces. [You can read our code conventions here](https://www.thecsharpacademy.com/article/58/code-conventions)🤓.

👪Oh and if possible could you please update your Discord username [in your profile](https://thecsharpacademy.com/dashboard/profile). 😁 Seems like it's '.myhos' and not 'myhos'

PR #573, opened 2025-09-08
Math game- All requirements + All challenges (- AI)

iDEaXANA:
Literally this https://www.youtube.com/watch?v=tuairRNAfZs

Will watch the official solution now :)

PR #590, opened 2025-10-26
math game :)

iDEaXANA:
I implemented the changes as per the old PR. Please see this for reference: https://github.com/TheCSharpAcademy/CodeReviews.Console.MathGame/pull/573

### DIFF
--- a/mathgame.iDEaXANA/Program.cs
+++ b/mathgame.iDEaXANA/Program.cs
@@ -1,7 +1,6 @@
 ﻿using System.Diagnostics;
-using System.Text;
 
-internal static class mathGame
+internal static class MathGame
 {
     private static readonly Random Rng = new Random();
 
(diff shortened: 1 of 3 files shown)


TheCSharpAcademy:
@iDEaXANA  Project approved! 😄✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster.

🔍️To be honest I don't have a lot of feedback, your code is clean and tidy and you follow good principles (SoC, DRY, etc). Keep up the good work. 

👨‍🏫One suggestion I'd give is to stay away from tuples when possible. Technically there's nothing wrong with using them, but they're ugly and hard to read. It'd be better to create a custom type here to abstract these properties: 

image

🗺️And do you mind adding a country in your profile as well? The dashboard and leaderboard look neater when there's a flag. If you're not comfortable with that you can submit a random country.

PR #623, opened 2025-12-17
Initial submission

ghost:
Thank you :)

### DIFF
--- a/MathGame.m1chael888/MathGame.slnx
+++ b/MathGame.m1chael888/MathGame.slnx
@@ -1,0 +1,3 @@
+<Solution>
+  <Project Path="MathGame/MathGame.csproj" />
+</Solution>
--- a/MathGame.m1chael888/MathGame/MathGame.csproj
+++ b/MathGame.m1chael888/MathGame/MathGame.csproj
@@ -1,0 +1,10 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net10.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+</Project>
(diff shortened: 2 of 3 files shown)


TheCSharpAcademy (changes requested):
@m1chael888 Thanks for submitting! 👏👏Your project is looking good, however one requirement has a bug. The history isn't saving at all 😔. Check this out:

image

🧼🫧Also, make sure you clean up remove unnecessary comments and unused usings, [as per our code conventions](https://thecsharpacademy.com/article/58/code-conventions).

🖥️Another suggestion: Clear the console more often. As you saw in the picture, it gets cluttered pretty quickly.

🔵Once you commit the changes you don't need to do a new PR, it will save and notify me.

ghost:
Thank you for the feedback :) I removed the unused using statement and implemented console clearing when returning to the menu or starting a new question. 

I may have misunderstood the 2nd rule which states "A game needs to have at least 5 questions.", which is why the game in your screenshot did not save ^^ I interpreted the rule to mean do not save a game with less than 5 questions. I have updated the logic to save any game with at least one question attempted. 

Thanks again :)

TheCSharpAcademy:
@m1chael888  Project approved!! 😄✅ It's a shame you didn't complete more challenges this time. 😟 It's always a good use of your time. But nothing stops you from revisiting the project later. 

🔍️To be honest I don't have a lot of feedback, your code is clean and tidy and you follow good principles (SoC, DRY, etc). Keep up the good work! The only thing I could say is that your implementation of the "5 questions" requirement could have been better. Ideally you'd want to loop 5 questions to the user after they chose a game. It's not very good UX to get the user to choose an operation after each question. 

🔍 Also, if each question is a new game, then save the question on each game! 😎 Not only you're saving the history after five games, but only in the correct answers, which makes the history not informative at all. The only thing the history ever says is: You got 5 points or more. Ideally it would be something like: 19/Dec/25 - Division Game - Points: 2 out of 5 - Time: 25". Or something like that.

🕵️‍♀️To finalize, have a look at our style guide since you still have dozens of unnecessary comments. 

PR #629, opened 2026-01-06
math game review

0lcm:


TheCSharpAcademy:
@0lcm  Project approved! 😄✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster.

**Feedback**:
🔍️The project is well-organized, you have one method for all games, which is great. One thing I would suggest is that you use type safety more often. There are too many magic strings in your app while you could have used enums for anything that has a limited amount of options. That makes the app less bug-prone and you have help from the compiler, which enhances your dev experience.

🔍️Using arrays is not necessarily bad, but List is much more flexible, since you don't have to know the number of items in advance. Queue<string[]> (in the picture below) is just messy. Much better to define a GameHistory type and use List<GameHistory>

image

🕵️‍♀️Also, have a look at our style guide since there are a few issues in your code. e.g. lots of unnecessary comments, as per picture below. [You can read our code conventions here](https://www.thecsharpacademy.com/article/58/code-conventions)🤓.

image

✋🏻Overal great work! Looking forward to seeing your next projects!

PR #630, opened 2026-01-06
TheCSharpAcademy Math Game Review

AntonOtto:


TheCSharpAcademy:
@AntonOtto  Project approved! 😄✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster.

**Feedback**
🔍️Your code is very well organized. Great work. 🤓. The first thing I'd say is that in your first menu, you didn't need the if-else at all. Just add a default case to your switch and let it handle it. You didn' t even need to parse it, just say : case "1", case "2" etc and the rest will be handled by the default case.

🔍️To make your randoms truly random, it would be better to create an instance on top, that would be used everywhere instead of  creating a new random on each game.  Random is seeded when you create it (based on the current time) and in .NET, if you call new Random() repeatedly in a short time, many of those instances end up with the same seed (or very similar seeds). If you play it enough times you'll see lots of repetition instead of true randomness 🥹

PR #634, opened 2026-01-11
project review

kaylubr:


mathGame.kaylubr/Program.cs
```diff
@@ -0,0 +1,336 @@
...
+        num1 = random.Next(50, 1001);
+        num2 = random.Next(50, 1001);
+      }
+    }
+
+
+    return [num1, num2];
+  }
+
+  public static string ValidateAnswer(string result)
```
TheCSharpAcademy: Input Validation shouldn't be handled with exceptions. Wrong inputs/answers are not exceptions 😆. Better to handle this with a loop. I'll let you investigate how to achieve this.

mathGame.kaylubr/Program.cs
```diff
@@ -0,0 +1,336 @@
...
+
+public enum MathGameDifficulty
+{
+  Normal,
+  Advanced
+}
+
+public class Program
+{
+  static List<dynamic> history = [];
```
TheCSharpAcademy: This isn't a requirement, but the earliest you learn this the better. One of C#'s main strengths is that it's type strong. I recommend you create a custom type instead of using dynamic. Then your collection would be something like List<GameHistory> history = new();

TheCSharpAcademy (changes requested):
@kaylubr Thanks for the submission! Your project is looking good but at the moment it doesn't meet this requirement:

image


image

TheCSharpAcademy:
@kaylubr Project approved! Well done for reworking your project.😄✅ And well done for completing some of the challenges! We recommend you always do that, as it pushes you and makes you learn faster.

**Feedback**
🔍️In the switch statement for your menu, you could used the default case to show a "wrong input" message. At the moment if I choose 9, for example, it simply reruns the loop without any message. Always think of making it really obvious for the user, with lots of messages telling them what's going on and what to do. 
<br>
image
<br>
🔍️You're only using two levels of indentation. While not wrong, the convention for .NET/C# devs is to use four.
<br>
🔍️Your code is well organized and follows DRY principles, except for the generation of numbers. You're repeating a lot of code to do that. Ideally you'd set variables (i.e. number ranges) that would change according to the difficult level.
<br>
image
<br>

🧑‍🏫Overall well done! 👏👏Looking forward to seeing your next projects!

PR #638, opened 2026-01-14
Challenges : DRY principles and random game

monsieur-raptor:


TheCSharpAcademy:
@monsieur-raptor Unfortunately we can't review projects after they have been approved, but it's great that you've enhanced it with your changes 😎. I suggest you keep it in your personal portfolio for reference.

PR #641, opened 2026-01-25
project completed

slimeswamps:


TheCSharpAcademy:
@slimeswamps  Project approved! 😄✅ It's a shame you didn't complete the challenges this time. 😟 It's always a good use of your time. But nothing stops you from revisiting the project later. 

**Feedback**
🔍️The biggest thing that comes to mind is that instead of using so many "if-else" you could have used a switch statement, which is much more elegant and less verbose. You could por example have handled this using the default case. I strongly recommend you look into this when creating menus for your console apps. 

🔍️Your list of results could show something more meaningful than just an integer. Example: 28/Jan/2026 - Divison Game - 3 Points. 

🔍️Make sure you don't leave unnecessary comments in the future or your project won't be approved. [You can read our code conventions here](https://www.thecsharpacademy.com/article/58/code-conventions)🤓.

PR #642, opened 2026-01-28
Project Finished

Ruksan12:
For review and upgrading


TheCSharpAcademy (changes requested):
@Ruksan12 Thanks for submitting the project! It's looking good, except that there's a bug, in your divisions if your range is 0, 99 sometimes this will happen and the app crashes: 
<br>
Image
<br>
Also this is not a requirement but it would be good if you'd handle the case where the user doesn't input an integer when answering a question. At the moment the app also crashes. Ideally you want to prevent this and show the user a meaningful error message instead of letting the app crash:
<br>
Image

Please fix the bug and optionally add validation so we can approve 😎

TheCSharpAcademy:
@Ruksan12  Project approved! 😄✅  It's a shame you didn't complete the challenges this time. 😟 It's always a good use of your time. But nothing stops you from revisiting the project later. 

🔍️In terms of feedback, the main thing: I highly recommend you learn about the [DRY Principle (Don't Repeat Yourself) ](https://www.thecsharpacademy.com/article/30006/dry-principle-csharp). It's the most important challenge in this project. Your code is neatly organized but there's a lot of repetition for generating the games. You should be able to have all games using just one method.

🔍️Also, consider clearing the console after each operation as it gets crowded quickly: 
<br>
image

🔍️At the moment you create a new Random() for each game. It's best practice to use a shared static Random or the generated numbers will be less random then expected.

☕🍵 If you think we're being helpful, [consider contributing with a donation](https://thecsharpacademy.com/contribute), any amount helps us keep our free service constantly improving.

PR #652, opened 2026-02-17
Review Requested

rcraig14:


TheCSharpAcademy:
@rcraig14  Project approved! 😄✅ It's a shame you didn't complete the challenges this time. 😟 It's always a good use of your time. But nothing stops you from revisiting the project later. 

**Feedback**:
🔍 As we can see in the picture below, the program crashes if I provide the wrong input in one of the menus. That's not very good user experience. You could put the menu in a loop and simply ask the user to input a valid option. Validation in general doesn't need to be handled with exceptions, since they are not ..... exceptions 😀
<br>
image
<br>

🔍Whenever developing console apps try to add more separation and clear the console more often (maybe after each operation?), as it's get cluttered pretty quickly :
<br>
image
<br>

🔍Wrapping a one-liner with a method doesn't really make a lot of sense. Why not just call console.writeline when needed? 
<br>
image
<br>

🔍There's no need to use recursion (i.e. a method that calls itself) when a loop can achieve the same result. Recursion is expensive. Essentially you're calling the same method over and over without returning. That **creates a new call stack frame** containing the method’s local variables, the return address (where to go back after it finishes), any parameters passed in, etc. If you use a loop the method is only called once at a time.
<br>
image
<br>
👪Oh and if possible join our Discord Community and add your username to your profile so you can be part of the community as well. 😁 And do you mind adding a country in your profile as well? The dashboard and leaderboard look neater when there's a flag. If you're not comfortable with that you can submit a random country.

☕🍵 If you think we're being helpful, [consider contributing with a donation](https://thecsharpacademy.com/contribute), any amount help us keep our free service constantly improving.

Overall great work and I'll see you in the next project!

PR #654, opened 2026-02-19
Code Review Request

leolin2810:


TheCSharpAcademy (changes requested):
@leolin2810 Thanks for submitting! The project looks good except that score is broken. I got all answers right but my score was 0:
<br>
Image
<br>
 Since we're here: A couple of things you could improve (not mandatory, but would be good).

🔍Scores at the moment show a block of "current settings". It would be better if it was a one liner per game: Date - Game Type, Score, etc.

🔍It's better to create one instance of Random and use it everywhere instead of creating one per operation. I'll let you research why. 

🔍For tracking the time of a game I suggest you look into the Stopwatch class, provided by the .NET class library, instead of: 

Image

TheCSharpAcademy:
@leolin2810 It's been two weeks since the suggested changes, so we're archiving this project. Feel free to reopen the PR and resubmit in the academy when you're ready. 🤓

PR #658, opened 2026-02-28
Project finished!

Ledana:
i tried and i know i could have done better. i believe it works but it is not pretty. not yet though. i will do better in the other one.

### DIFF
--- a/MathGame/MathGame.csproj
+++ b/MathGame/MathGame.csproj
@@ -1,0 +1,10 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+</Project>
--- a/MathGame/MathGame.sln
+++ b/MathGame/MathGame.sln
@@ -1,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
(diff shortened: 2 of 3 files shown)


TheCSharpAcademy:
@Ledana Project approved!  😄✅ It's a shame you didn't complete the challenges this time. 😟 It's always a good use of your time. But nothing stops you from revisiting the project later. 

In terms of feedback, just a couple of things: 
🔍️I highly recommend you learn about the [DRY Principle (Don't Repeat Yourself) ](https://www.thecsharpacademy.com/article/30006/dry-principle-csharp). It's the most important challenge in this project. 
Your code is neatly organized but Cases 1 through 4 in **RunGame(Random, List<string>, int)** are nearly identical. The only difference is the operator character and the arithmetic operation. 80 lines of duplicated logic could be reduced to a single method. I suggest you try to refactor it as a learning exercise.

🔍On line 72 the timer starts with **timeStart = DateTime.UtcNow;** but in line 222 it ends with **timeStart = DateTime.UtcNow;**. These use different time zones. DateTime.UtcNow returns UTC time while DateTime.Now returns local time. The elapsed time calculation (timeFinish - timeStart).Seconds will be incorrect by the UTC offset of the user's timezone (e.g., off by hours in most locales). Both should use the same source or use **System.Diagnostics.Stopwatch** for measuring elapsed time.

🕵️‍♀️Also, have a look at our style guide since there are a few issues in your code. e.g. lots of unused usings, and unnecessary comments. [You can read our code conventions here](https://www.thecsharpacademy.com/article/58/code-conventions)🤓.

👪Oh and if possible join our Discord Community and add your username to your profile so you can be part of the community as well. 😁

Overall well done!✋🏻Looking forward to seeing your next projects!

Ledana:
Thank you so much! i will keep in mind every advice for future projects. and return to the math game to improve it.

PR #659, opened 2026-02-28
MathGame Submission

Tenebris-06:


TheCSharpAcademy:
@Tenebris-06  Project approved! 😄✅ It's a shame you didn't complete the challenges this time. 😟 It's always a good use of your time. But nothing stops you from revisiting the project later. I like that you have a ProblemGenerator that handles all games. Well done. 💪

In terms of feedback, just a few things to think about for the next projects: 
🔍️At the moment your games menu have flawed validation. I can input an option for a non-existant game and this is what happens: 
image
<br> Instead of checking for integers you could simple check for the correct characters in a switch statement and handle everything else with a default case. Additionally if I input a string the app breaks:
image
<br>
🔍️In your Problem class, you could have used auto-properties. They reduce the boiler plate and your file would be down to 10 lines instead of 60.

🔍️It's good that you're dividing your code in different classes and files. The next step is to remove everything from Program.cs and use it merely as an entry-point.

Tenebris-06:
Thank you for the comprehensive feedback!
Will try to pay attention to everything mentioned.

PR #664, opened 2026-03-04
Complete math game project

royeeet:
Completed math game project, including base requirements and challenge steps!


TheCSharpAcademy (changes requested):
@royeeet Thanks for the submission. Unfortunately I couldn't review since you're using .NET framework v.4.8. 😥

While it still has support this is considered a legacy technology. I don't even have it installed in my computer. Please update your project to at least .NET 8 before we can review. 😀

TheCSharpAcademy:
@royeeet Project approved! 😄✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster.

🔍️To be honest I don't have a lot of feedback, you ticked all boxes. I particularly like you're using only one method for all functions.

🔍️The only thing I found is that it doesn't make sense to throw an exception and let the app crash upon an incorrect input:
image
<br>
You could have handled wrong inputs in the difficulty menu the same way you handle the validation of the answers. Present a message telling the user what they did wrong and let them try again.

Overall great work✋🏻Looking forward to seeing your next projects!

PR #666, opened 2026-03-06
Added project. Final to submit

ProximaEx:


TheCSharpAcademy:
@ProximaEx Project approved! 😄✅ And well done for completing some of challenges! I recommend you always do that, as it pushes you and makes you learn faster. I love the UI. It's very nice to let the users move through the menu using arrows. Very original. One of the most interesting I've seen in a console app without the use of libraries. I also really like that you let the users choose the highest number. Keep up doing creative things like that! 👏

🔍️In terms of feedback, just a couple of things: I highly recommend you learn about the [DRY Principle (Don't Repeat Yourself) ](https://www.thecsharpacademy.com/article/30006/dry-principle-csharp). It's the most important challenge in this project. Your code is neatly organized but there's a lot of repetition for the games in your switch statement. You should be able to have all games using the same code and just different variables. 

🔍️ It's best practice to use one instance of Random for your app and use it everywhere, instead of creating a new instance for each round. 

🕵️‍♀️Also, have a look at our style guide since your code is littered with unnecessary comments. [You can read our code conventions here](https://www.thecsharpacademy.com/article/58/code-conventions)🤓.

PR #669, opened 2026-03-13
All code changes

stevenwardlaw:


TheCSharpAcademy:
@stevenwardlaw  Project approved! 😄✅ And well done for completing some of the challenges! I recommend you always do that, as it pushes you and makes you learn faster.

**Feedback**
🔍️You're allowing users to see a history of games, but there's no option in the menu for that. This was a requirement, but we will let it pass this time 😀. Also your history is purely an integer, which is not very useful to the user. Ideally each record should contain the date, game type, score, etc. 

🕵️‍♀️Also, have a look at our style guide since there are too many unnecessary comments in your code. It's ok to use AI to help you, but don't leave their garbage comments unless they explain something that can't be inferred by the code. [You can read our code conventions here](https://www.thecsharpacademy.com/article/58/code-conventions)🤓: 

image

PR #672, opened 2026-03-21
Done

nice-Code-90:
fixes added


nice-Code-90:
Codacy gave some false positive here.

TheCSharpAcademy:
@nice-Code-90  Project approved! 😄✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster.

**Feedback**

🔍️Your code is well organized. I like that you used appsettings for configurations. Not many people do that in their first project. You also elegantly added your own modifications to the original tutorial's code, which is always a good thing 😎

🔍️ I found a bug where after providing the wrong input for a question the app breaks. I recommend you always use TryParse instead of parse to avoid these types of headaches: 

image
<br>

PR #676, opened 2026-03-26
Finished the Console.MathGame that was fun.

JakePorter05:
Thank you for the exercise. Please review. I am open to criticism.

### DIFF
--- a/Console.MathGame.JakePorter05/Console.MathGame.JakePorter05.slnx
+++ b/Console.MathGame.JakePorter05/Console.MathGame.JakePorter05.slnx
@@ -1,0 +1,3 @@
+<Solution>
+  <Project Path="Console.MathGame.JakePorter05/Console.MathGame.JakePorter05.csproj" />
+</Solution>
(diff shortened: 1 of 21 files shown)


TheCSharpAcademy:
@JakePorter05  Project approved! 😄✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster. And I really like the Harry Potter twist 🧙‍♂️. Also it's good to see entity framework in this first project, although it wasn't a requirement. 😀

🔍️In terms of feedback, just one thing: For anything that has a limited amount of options and it's not set by the user,  it would be better to use enums than magic strings. This way you'd get help from the compiler, which decreases errors. That's the case for Grades, Houses, and GameType.

🔍Your year validation message says “between 1 and 7,” but the code only checks whether the input is an integer. That means values like 0, 12, or -3 would still be accepted. I didn't understand what year was the first time around and put 2025, which means that I ended up in an endless game with 2000+ questions.

JakePorter05:
I’ve gotten into the habit of using string constants for Pickers and other UI elements so I don’t have to write a conversion method. Because of that, I tend to forget about enums entirely. It’s good to know that using enums would actually be the better approach.
I’ll also fix the infinite game issue, since I’m planning to save this project to my GitHub as part of building out a portfolio.

PR #691, opened 2026-05-03
Added my finished project

Hacker-735:


TheCSharpAcademy (changes requested):
@Hacker-735 Thanks for submitting! It's looking good but there are a few things that need improvement:

- Don't make the user choose an operation for each question. Let them choose an operation then ask the set amount of times.
- The game crashes after it's finished. This isn't very good user experience. After the five questions, loop it back to the menu.
- If I type something random it says I typed 0, which is not true.

Image

Once you make the changes just commit and I'll be notified. No need to create another PR :)

Hacker-735:
After implementing the changes you requested I pressed Push in visual studio here, [cid:09454ee8-a380-42bc-9a09-54fa42263b22]

Do I need to do anything else on github? I can't see the options that were available last time anymore.
________________________________

TheCSharpAcademy:
@Hacker-735  Project approved! Thanks for making the requested changes😄✅ It's a shame you didn't complete the challenges this time. 😟 It's always a good use of your time. But nothing stops you from revisiting the project later. 

🔍️For feedback, just a couple of things. I saw you created a nested loop (a do while inside another one), to keep the game running. While this is not wrong, if you had a third menu (let's say to handle difficulty), then you'd have a nested loop 3 levels deep and it would become very hard to debug. If possible avoid many layers of loops or at least put the inner loop inside it's own method so it's easier to debug.

🔍️As you can see below, at the moment the history is not very readable. While you're correctly storing records in a list as requested, ideally you'd want a cleaner history per game instead of per question. Example:
05-May-26 09:30 - Division - 4/5. 
image

PR #692, opened 2026-05-06
Aaron Kagan Math Game

aaronkagan:


TheCSharpAcademy:
@aaronkagan  Project approved! 😄✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster.

🔍️In terms of feedback, just a few minor things. First, there's a bit of repetition in your question generation. In the picture below, for example, you could have a single block with the logic after the switch statement. The ranges would be set by variables. And the switch cases would only assign those variables depending on the difficulty. Also, the indentation of the switch is a bit messed up; you have too many levels from line 132:
image

🔍️For better readability I recommend you clear the console after each operation, as it gets cluttered really quickly:
image

🔍️Nice use of OOP, it looks like your hard work is bearing fruits and you're getting more comfortable with custom classes and objects. FYI by convention in C# classes are generally separated into their own files. That will be enforced from the 4th project (coding tracker)

🔍️Also in the picture above, when the input doesn't match one of the menu cases, it fails silently. You want a validation error message similar to what you do when validating answers for integers.
image

🔍️When committing your code make sure you only commit files that are absolutely necessary to run the project. Both the .idea folders didn't need to be there:
image

PR #698, opened 2026-05-21
My MathGame Submission

awerner9:


TheCSharpAcademy:
@awerner9  Project approved! 😄✅ It's a shame you didn't complete the challenges this time. 😟 It's always a good use of your time. But nothing stops you from revisiting the project later. 

🥋And congratulations! You got your green belt!! 🟩🟩 Roll-up your sleeves, it's just the start 😂

**Feedback**
🔍️I highly recommend you learn about the [DRY Principle (Don't Repeat Yourself) ](https://www.thecsharpacademy.com/article/30006/dry-principle-csharp). It's the most important challenge in this project. Your code is neatly organized but there's a lot of repetition for the games. You should be able to have all games using just one method.

🔍️Your menu is a bit verbose. A switch statement would be way more elegant. It's better suited for conditionals with multiple candidates.<br>
image

🔍️When using console UIs, always make sure you clear the screen often or it gets cluttered quickly: <br>image

🔍️Use **int.TryParse** instead of **int.Parse**: Right now, if the user types "abc" as an answer, the app will crash. This is the biggest improvement I recommend.

🔍️Avoid calling **Menu(results)** recursively: Each game method calls the menu again. It works for small console apps, but over time this can build up unnecessary call stack depth. A while loop in Main is a cleaner structure.

Overall good work!✋🏻Looking forward to seeing your next projects!

PR #707, opened 2026-06-24
project finished!

abahm00:


TheCSharpAcademy:
@abahm0 Project approved! 😄✅ Great work adding a timer with Stopwatch. That is a nice extra feature and gives the game history more useful information.

**Feedback**
🔍️Random random = new Random(); is created every question. Prefer one shared Random instance so randomness is more reliable.

🔍️CheckIfAnswerCorrect has repeated logic for each operation. You could calculate the expected answer first, then compare once.

🔍️The question prompt says “between 5 and 10,” but some error messages say “between 1 and 10.”, as per picture below. Keep those consistent.

image

🔍️Consider adding an explicit Exit option to the main menu instead of only checking for e after each action.

PR #709, opened 2026-06-28
MathGame Project finished

rakenkode:
Looking forward to getting real feedback from a professional like you. Thanks in advance.

### DIFF
--- a/Rakenkode.MathGame/.idea/.idea.Rakenkode.MathGame/.idea/.gitignore
+++ b/Rakenkode.MathGame/.idea/.idea.Rakenkode.MathGame/.idea/.gitignore
@@ -1,0 +1,15 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Rider ignored files
+/contentModel.xml
+/projectSettingsUpdater.xml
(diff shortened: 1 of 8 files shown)


TheCSharpAcademy:
@rakenkode  Project approved! 😄✅ And well done for completing some of the challenges! I recommend you always do that, as it pushes you and makes you learn faster. You also did a great work with the UI and UX, very smooth! 👏

🔍️Avoid magic numbers. Although _numberOfQuestions exists, the value 5 is still hardcoded in several places (if (_questionNumber < 5), "GAME X of 5", if (_questionNumber == 5)). Use _numberOfQuestions consistently instead.

🔍️Question contains two overloads of CreateProblems(), but only the GameType version is required. The CreateProblems(int mode) overload appears to be unused and should be removed to keep the codebase clean.

🔍️Every Question creates its own Random instance, and Thread.Sleep() is used to try to avoid duplicate seeds. This is generally considered a bad practice. Instead, use a single shared Random instance (or Random.Shared in modern .NET) and remove the Thread.Sleep() calls entirely.

PR #712, opened 2026-06-29
MathGame solutioon

matejadb:
Hello, 
This is my solution to the C# Academy MathGame project.

I had trouble importing my already created repository into the fork so I had to create a new project, hence there is only one commit 😄

### DIFF
--- a/MathGame.matejadb/History.cs
+++ b/MathGame.matejadb/History.cs
@@ -1,0 +1,7 @@
+﻿namespace MathGame.matejadb;
+
+internal class History
+{
+    public int GameNumber { get; set; }
+    public int Points { get; set; }
+}
--- a/MathGame.matejadb/History.cs
+++ b/MathGame.matejadb/History.cs
@@ -3,5 +3,6 @@
 internal class History
 {
     public int GameNumber { get; set; }
+    public string? Answer { get; set; }
     public int Points { get; set; }
 }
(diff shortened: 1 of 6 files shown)


TheCSharpAcademy (changes requested):
@matejadb Thanks for submitting! Unfortunately, there's a core requirement missing as there's no menu to choose an operation:

Image

matejadb:
> @matejadb Thanks for submitting! Unfortunately, there's a core requirement missing as there's no menu to choose an operation:
> Image

My bad, misunderstood the requirement. It should be implemented now!

TheCSharpAcademy:
@matejadb  Project approved! 😄✅ It's a shame you didn't complete the challenges this time, but nothing stops you from coming back later to complete them. We strongly suggest you do 😁

**Feedback**
🔍️**Avoid recursive menu calls:** Right now GameMenu(), calls SelectOption(), and many options call GameMenu() again. Also GenerateQuestion(selectedOperation) recurse many times. This works for small apps, but repeated recursion can build up the call stack. 

🔍️Your numbers are generated as whole numbers and answers are checked as integers, so Question can use int:
image

🔍Currently each round is saved as a separate history item, but it says GameNumber, which makes it look like one entry per game.You could store one history entry per completed game:
```
internal class History
{
    public int GameNumber { get; set; }
    public string Operation { get; set; } = string.Empty;
    public int Score { get; set; }
    public int TotalRounds { get; set; }
    public DateTime DatePlayed { get; set; }
}
```

☕If you like our roadmap, please consider [buying as a coffee](https://thecsharpacademy.com/Contribute). We appreciate your help 🙂

PR #717, opened 2026-07-08
Finished project

dewsev:
Please be honest about the code.

My IDE shows that the switch expressions are not exhaustive, any tips?
I didn't want to throw exceptions in the discard arm, because that would require more exception handling, which would make the code a bit too complicated for this little project (that's what I think at least) but maybe that's what I should be doing?

Looking forward for the feedback!
Thank you in advance.


TheCSharpAcademy:
@dewsev Project approved! 😄✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster. Your project is clean, tidy, concise and to the point. And you're using only one method for all operations, great work!

**Feedback**
🔍️Dispose the timer. Currently you create the timer like this:
```
_ = new Timer(...);
```
Because the reference is discarded, the timer continues running until the application exits. A better approach would be:
```
using Timer timer = new(...);
```
Otherwise the timer continues incrementing in the background.

🔍️Simplify boolean returns. For example:
```
if (expected == answer)
{
    return true;
}

return false;
```
can simply become
```
return expected == answer;
```
Small improvement, but you'll see this style throughout modern C#.

🔍Adding the date and time a game was played would make the history feature more informative.

🔍Tuples are messy. Instead of List<(OperationType, DifficultyLevel, int, int)> you could create:
```
class GameResult
{
    public OperationType Operation { get; init; }
    public DifficultyLevel Difficulty { get; init; }
    public TimeSpan Duration { get; init; }
    public int Score { get; init; }
}
```
Named properties tend to make code easier to read than tuple indices (Item1, Item2, etc.), especially as projects grow.

☕If you like our roadmap, please consider [buying as a coffee](https://thecsharpacademy.com/Contribute). We appreciate your help 🙂

👪Oh and if possible join our Discord Community and add your username [to your profile](https://thecsharpacademy.com/dashboard/profile) so you can be part of the community as well. 😁

dewsev:
Thank you for reviewing my project.
I have made all the changes you mentioned and can't wait to start the next project!

PR #721, opened 2026-07-11
Project is ready

Solomonlol:


TheCSharpAcademy (changes requested):
@Solomonlol Thanks for submitting! This project needs a bit of work before we can approve. Could you please fix the issues below?

🔍The main menu runs inside an infinite loop, but there is no option to exit the application. Please add an explicit exit option so the user does not have to close the console window manually.

🔍The application does not currently store or display previous games. A completed game should normally record information such as: Date played, Operation, Score

🔍Correct and incorrect answers are displayed, but the application does not keep a total score. Consider incrementing a score for every correct answer and showing the result when the game ends.

🔍The user can enter zero or a negative number of questions. In that situation, the game ends immediately. Validate that the number is greater than zero before starting the game.

🔍When the player enters text instead of a number, the question counter is still incremented. It would provide a better experience to keep asking until a valid integer is entered.

Solomonlol:
Okay, thanks for checking. I'll fix everything and try again :)

TheCSharpAcademy (changes requested):
@Solomonlol Thanks for making the changes.

I'm really sorry but I don't speak your language (Belarusian?). Do you mind translating your app to English?

Image

Solomonlol:
Done

TheCSharpAcademy:
@Solomonlol Project approved! 😄✅

Nice job! You covered all the core requirements and completed several of the optional challenges as well. I liked that you added difficulty levels, a timer, and game history. Splitting the application into multiple classes instead of putting everything in Program.cs also makes the project much easier to follow.

**Feedback**

🕵️‍♀️Have a look at our style guide since there are a few issues in your code. e.g. lots of unused usings, as per picture below. [You can read our code conventions here](https://www.thecsharpacademy.com/article/58/code-conventions)🤓.

image

🔍️ MathGame.Game() contains a lot of duplicated code. The only thing that changes between each switch case is the calculation itself, while the logic for checking the answer, updating the score, and displaying the result is identical. You could calculate the expected answer first and then perform the comparison once, making the method much shorter and easier to maintain.

🔍️ A new Random instance is created every time GenerateNumbers() is called. It's generally better to create a single Random object and reuse it throughout the lifetime of the application.

🔍️ I noticed that GameHistory() calls MainMenu() after displaying the history. Since MainMenu() is already running in a loop, this creates unnecessary recursion. Simply returning to the existing loop would keep the navigation simpler and avoid growing the call stack over time.

🔍️ GameResults only stores the operation and final score. It would be even more useful if the history also included information such as the difficulty level and the time taken to complete the game, since you're already calculating the elapsed time.

☕ If you like our roadmap, please consider buying us a coffee ☕🙂
https://thecsharpacademy.com/Contribute

Overall, great work! ✋🏻

PR #723, opened 2026-07-12
My implementation of the MathGame project.

BudgetKratos:
I took the advice below a bit too literally, this was my **_third_** attempt at the Math Game.

> If this is your first project, I recommend doing it _twice_. You'll be surprised at how much your retention increases, and how many gaps were left in the first time.

I'm sure if I were left to my own devices, I would still be obsessing over the tiny details, trying to make everything perfect. 

It's a habit I should kick or, at the very least, channel into something else!

### DIFF
--- a/MathGame.BudgetKratos/.editorconfig
+++ b/MathGame.BudgetKratos/.editorconfig
@@ -1,0 +1,208 @@
+# editorconfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Default settings:
+# A newline ending every file
+# Use 4 spaces as indentation
+[*]
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+
+# Specify UTF-8 without byte-order mark
+[*.{csproj,locproj,nativeproj,proj,resx,slnx,vbproj}]
+charset = utf-8
+
+# Generated code
+[*{_AssemblyInfo.cs,.notsupported.cs,AsmOffsets.cs}]
+generated_code = true
+
+# C# files
+[*.cs]
+# New line preferences
+csharp_new_line_before_open_brace = all
+csharp_new_line_before_else = true
+csharp_new_line_before_catch = true
+csharp_new_line_before_finally = true
+csharp_new_line_before_members_in_object_initializers = true
+csharp_new_line_before_members_in_anonymous_types = true
+csharp_new_line_between_query_expression_clauses = true
+
+# Indentation preferences
+csharp_indent_block_contents = true
+csharp_indent_braces = false
+csharp_indent_case_contents = true
+csharp_indent_case_contents_when_block = false
+csharp_indent_switch_labels = true
+csharp_indent_labels = one_less_than_current
+
+# Modifier preferences
+csharp_preferred_modifier_order = public,private,protected,internal,file,static,extern,new,virtual,abstract,sealed,override,readonly,unsafe,required,volatile,async:suggestion
+
+# avoid this. unless absolutely necessary
+dotnet_style_qualification_for_field = false:suggestion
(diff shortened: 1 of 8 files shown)


TheCSharpAcademy (changes requested):
@BudgetKratos Thanks for submitting! There's a bug blocking in the history, which blocks the approval, could you please check?

⚠️ The score-history flow needs fixing.
When the player chooses not to play again, Operate() calls Menu() before reaching ScoreCard.Add(...). Consequently, the completed result is not added to the score history until the nested menu eventually returns. Repeated navigation also creates an unnecessary chain of recursive method calls.

Instead of calling Menu() from inside Operate(), let Operate() finish naturally:
```
if (!retry)
{
    ScoreCard.Add($"{CurrentDifficulty} {actualOperation} {score}");

    DisplayMessage("Press any key to go back to the previous menu.", displayColor);
    Console.ReadKey();
    return;
}
```

Upon fixing, just commit your changes, no need for a new PR 😁

BudgetKratos:
Good catch, thank you ^_^

I've also slightly updated the code so that every retry counts as a separate attempt (as I had originally intended).

TheCSharpAcademy:
@BudgetKratos Project approved! 😄✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster. Your project is clean, tidy, concise and to the point. And you're using only one method for all operations, great work!

---

## **Feedback**

💖You showed good use of inheritance, separation of concerns and nice helper methods. 👏👏

---

🔍️1. Avoid storing scores as strings. This is probably my biggest suggestion. Currently:
```
ScoreCard.Add($"{CurrentDifficulty} {actualOperation} {score}");
```
Later you split it again.
```
string[] subStrings = score.Split();
```
Instead, create a small model.

class Score
{
    public string Difficulty { get; set; }
    public char Operation { get; set; }
    public int Points { get; set; }
}

Then
```
List<Score>
```
This removes parsing entirely and is much more object-oriented.

---

🔍️ Avoid comparing enum names as strings

Instead of
```
difficulty.ToString() is "Medium"
```
compare the enum directly.
```
difficulty == DifficultyChoices.Medium
```
or even better, use a switch expression.
```
return difficulty switch
{
    DifficultyChoices.Easy => SetBounds(1,20),
    DifficultyChoices.Medium => SetBounds(21,50),
    ...
};
```
This is safer, faster and much more idiomatic C#.

---

🔍️ Reduce repeated switch logic. The operation selection contains several repeated patterns. For example
```
case GameChoices.Addition:
    Operate('+', "DarkOrange");
```
A dictionary or small configuration object could map operation colour and title to reduce maintenance if you ever add more games.

---

🔍Random can be non-nullable: Since it's initialized immediately
```
protected Random? RandomNumber { get; } = new();
```
the nullable annotation isn't necessary.

---
🔍CheckAnswer can be simplified. A switch expression makes it shorter, cleaner and easier to extend.
```
return operation switch
{
    '+' => answer == operands.Item1 + operands.Item2,
    '-' => answer == operands.Item1 - operands.Item2,
    '*' => answer == operands.Item1 * operands.Item2,
    '/' => answer == operands.Item1 / operands.Item2,
    _ => false
};
```

---

🗺️ I noticed that the only course you've completed so far is **Introduction to Object-Oriented Programming**. I'd highly recommend [taking our **LINQ Fundamentals** course next](https://thecsharpacademy.com/course/8/article/1/500249/False). LINQ is one of the most important tools in modern C# and becoming comfortable with it will make your code much cleaner and more expressive.

---

☕ If you like our roadmap, please consider [buying us a coffee](https://thecsharpacademy.com/Contribute). We appreciate your help 🙂

---

Overall great work! ✋🏻 Looking forward to seeing your next projects!

PR #726, opened 2026-07-20
MathGame.Louis-dub

Louis-dub:


TheCSharpAcademy:
@Louis-dub Project approved! 😄✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster. Your project is clean, tidy, concise and to the point. And you're using only one method for all operations, great work!

---

## **Feedback**

🔍️ I highly recommend you learn about the [DRY Principle (Don't Repeat Yourself)](https://www.thecsharpacademy.com/article/30006/dry-principle-csharp). There's a lot of repetition for the games. The code is almost identical and even though you showed a good use of inheritance, it could be simplified a lot.

---

🔍️ The history collection is globally mutable:

```csharp
internal static List<Operation> Operations = [];
```

Consider exposing it as a property and preventing it from being replaced:

```csharp
internal static List<Operation> Operations { get; } = [];
```

For stronger encapsulation, the class could expose methods such as `Add()` and `GetAll()` instead of exposing the list directly.

---

🔍️ Each question currently creates a new `Random` instance. Modern .NET handles rapid `Random` construction better than older versions, but reusing `Random.Shared` would still be simpler. This also avoids repeatedly constructing random generators.

```csharp
int n1 = Random.Shared.Next(1, mode);
```
---

🗺️ I noticed that the only course you've completed so far is Introduction to Object-Oriented Programming. I'd highly recommend [taking our LINQ Fundamentals course next](https://www.thecsharpacademy.com/). LINQ is one of the most important tools in modern C# and becoming comfortable with it will make your code much cleaner and more expressive.

---

👪 Oh and if possible join our Discord Community and add your username [to your profile](https://thecsharpacademy.com/dashboard/profile) so you can be part of the community as well. 😁 And do you mind adding a country in your profile as well? The dashboard and leaderboard look neater when there's a flag. If you're not comfortable with that you can submit a random country.

---

☕ If you like our roadmap, please consider [buying us a coffee](https://thecsharpacademy.com/Contribute). We appreciate your help 🙂

---

Overall great work! ✋🏻 Looking forward to seeing your next projects!

PR #733, opened 2026-08-02
Thanks for your great work!

MichaelHelgesen:


TheCSharpAcademy:
@MichaelHelgesen Project approved! 😄✅ It's a shame you didn't complete the challenges this time. 😟 It's always a good use of your time. But nothing stops you from revisiting the project later. 

**Feedback**
---
🔍️ The code in the picture below works but it's virtually unreadable. It would be better to handle the menu with a switch statement and let the default case handle invalid inputs.

image
<br>
Also if you do need to chain statements like this make it more readable by stacking them:

image

---

🔍️It's not good practice to handle input validation with Exceptions. Exceptions are for things that happen exceptionally and aren't expected. Input errors are expected. Better to check the input in a loop with a friendly message. In this case you could use int.TryParse() so instead of erroring out you would have to handle the null case.

---

🔍Your methods are doing too much. As an example, StartGame generate numbers, prints questions, collects input, evaluate responses, etc... It's ok to have a method that orchestrates things but this method could call other methods that do part of the job. That's particularly important when there are nested loops. It becomes very hard to debug.

---

🗺️ I noticed that you've only completed the first lesson of the **Introduction to Object-Oriented Programming** course. I'd highly recommend finishing it so you can better understand how to split your project in methods and classes and abide to the  separation of concerns principle. 😊

---

☕If you like our roadmap, please consider [buying us a coffee](https://thecsharpacademy.com/Contribute). We appreciate your help 🙂

PR #745, opened 2026-08-20
Finished Project!

AaronjMurphy:


TheCSharpAcademy (commented):
@AaronjMurphy Thanks for the submission! 🙂

It looks like the `.csproj` file is missing from the repository. Please include it in your submission, as without the project file we aren't able to build and run the project for review.

Once you've added and pushed the `.csproj` file, we can continue with the review. 👍

TheCSharpAcademy (changes requested):
@AaronjMurphy Thanks for the submission! 🙂

It looks like the `.csproj` file is missing from the repository. Please include it in your submission, as without the project file we aren't able to build and run the project for review.

Once you've added and pushed the `.csproj` file, we can continue with the review. 👍

PR #754, opened 2026-08-26
Finished

AbooSalh:


TheCSharpAcademy (commented):
@AbooSalh Thanks for submitting! This project needs a bit of work. Here are the missing requirements:

❌Ask math questions and award points for correct answers — The app asks questions and checks answers correctly through IsCorrect(), which is a good start. However, there is currently no score/point system. A correct answer needs to add a point.
❌ At least 5 questions per game — PlayGame() currently asks the player to select and answer one question, after which the game ends. A game needs at least five questions.
❌ The game only have four hard-coded questions. The flow instead needs to be: User chooses an operation, then five questions are randomly generated, so each game is different. 
❌ Menu to choose an operation — The current menu lets the user choose Play, View History, or Exit, and then choose an individual pre-written question. The player should instead be able to choose a mathematical operation such as addition, subtraction, multiplication, or division.

💁‍♂️If you have any questions, the best way to get help is [in our Discord](https://discord.com/invite/aDMDET8ywB), in the #get-help-with-projects channel. 😊

PR #757, opened 2026-08-28
Math Game finished!

Brandonscodedata:


TheCSharpAcademy (changes requested):
@Brandonscodedata Thanks for submitting. It's looking good except that there's no option **in the menu** for users to see the history. That is a requirement of the project.

Image

image

Additionally, it would be a better user experience if the users didn't have to choose the operation for every question. Instead give them the chosen number of questions for the chosen operations. Then in the history store the result of each game, **not each question**.

Once you make the changes, just commit them. A new Pull Request is not necessary.

PR #762, opened 2026-08-31
My finished challenge

lucascorreia-tech:


TheCSharpAcademy (changes requested):
@lucascorreia-tech Thanks for submitting your project. There are a couple of problems that you need to fix before we can review. First there's a bug where the app crashes when it tries to divide by zero:

Image

Additionally your app doesn't let the users choose the operation, which is a requirement of the project: 

Image

Let me know if you need help, sou brasileiro tambem. 😁

lucascorreia-tech (commented):
First of all, thank you for the reply. Because of that error, I reviewed the program's entire logic and refactored a large part of the code. Could you please take another look at it? VAMO Brazil! kkk

PR #764, opened 2026-09-03
submission for review

spheritual:


TheCSharpAcademy (changes requested):
@lucidest Thanks for submitting! There is one requirement not being met. Could you please fix it before we can finish the review? 

❌ A game currently contains only one question. Selecting Addition, Subtraction, Multiplication, or Division calls the corresponding method once and immediately returns to the menu. The requirement is that one game must have at least five questions. 

💠You want a "game" to be a group of questions. So in the history you'd show something like: Date - Game Type - Score

PR #776, opened 2026-09-24
Math game in CLI with challenges

Boudiz:


TheCSharpAcademy:
@Boudiz  Project approved! 😄✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster. Your project is clean, tidy, concise and to the point. And you're using only one method for all operations, great work!

🥋You still need to submit your [Freecodecamp certificate](https://thecsharpacademy.com/project/75/freecodecamp-certification) and read a few articles before you start earning belts 😊

**Feedback**
🔍️ There is one edge case worth revisiting there. After the timeout you do:
```
await inputTask;
```
Because Console.ReadLine() itself isn't actually cancelled by the token, the application still waits for the user to press Enter before continuing. Your comment acknowledges this, so it appears intentional, but it means the "20 second limit" doesn't immediately advance to the next question.

☕If you like our roadmap, please consider [buying us a coffee](https://thecsharpacademy.com/Contribute). We appreciate your help 🙂

👪Oh and if possible join our Discord Community and add your username [to your profile](https://thecsharpacademy.com/dashboard/profile) so you can be part of the community as well. 😁

Boudiz:
Thank you for the review,
The await inputTask is a workaround because if the task is cancelled and we don't wait for ENTER press, the Console.ReadLine is not canceled (I couldn't find a way to cancel it). And the next ReadLine of the next question would get queued after the ReadLine of the inputTask, even if the inputTask itself was canceled.

